PR #33, opened 2026-05-18
Adds an in-app language switcher (Chinese / English) to the Settings screen.

FunLay123:
The original app had no way to change language from within the app itself.

- `LocaleHelper` — stores chosen language in SharedPreferences, wraps
  `Context` via `createConfigurationContext()` (no deprecated APIs);
  auto-detects system locale on first launch
- `I18n.kt` — `tr()` helper for non-Compose code (ViewModels, Services)
- `AbkApplication` / `MainActivity` — override `attachBaseContext` so
  the saved locale applies on every cold start
- Settings UI — Language section between Navigation and Theme groups
  with 中文 / English buttons; selecting one calls `Activity.recreate()`
  for an immediate switch
  
  P.S. abk-manager-cert.env can be ignored


xingguangcuican6666:
plz review your code, there has some check failed. Thanks.

xingguangcuican6666:
OK, plz wait for my test. Thanks.

xingguangcuican6666:
It's seem has some problem when the language switching to English, Some text was still keeping Chinese. Could you fix it?
Screenshot_2026-05-19-12-23-47-352_com abk kernem
Screenshot_2026-05-19-12-23-43-193_com abk kernem
Screenshot_2026-05-19-12-23-11-632_com abk kernem

app/signing/abk-manager-cert.env
```diff
@@ -1,7 +1,5 @@
 # Official ABK release signing certificate metadata.
 # This is public certificate metadata, not a private signing key.
-# Keep this file in the repository so forked kernel builds can trust the
-# official ABK manager APK without requiring repository secrets.
 ABK_MANAGER_PACKAGE=com.abk.kernel
-ABK_MANAGER_CERT_SIZE=1407
-ABK_MANAGER_CERT_SHA256=34e5e843952277759603cd0f949770b24c868530d80d7baeff08776a7e132b16
+ABK_MANAGER_CERT_SIZE=838
+ABK_MANAGER_CERT_SHA256=c46ece2acbc92132f7d86967fd397e6629429c47dec938d4a98509dd9ff48e09
```
xingguangcuican6666: Signed info shouldn't commit at pr

xingguangcuican6666 (changes requested):

FunLay123:
thats not the final code, just testing something

prpjzz:
> The original app had no way to change language from within the app itself.

My design approach is to use the system's built-in App Language settings as the language setting, instead of having to integrate a language switching menu myself.

But in some ROM, especially HyperOS, i don't know how to go this settings and I must use intent to go to settings.

FunLay123:
## Update — fixes for the Chinese-text-after-switch bug

Pushed an update addressing your test report.

### 1. Fix for "some text still keeping Chinese" after switching to English

The strings on your screenshots were hardcoded in non-Compose code (`RootUtils`, `GitHubRepository`, `BuildProgressUtils`, `MainViewModel`). They've been extracted to `strings.xml` and replaced with `tr()` calls so they now follow the locale set by `LocaleHelper`:

| File | What |
|---|---|
| `RootUtils.kt` | Diagnostic messages (`ru_diag_compat_shell_only`, `ru_no_manager_interface`, `ru_diag_native_not_manager`, etc.) — all the warnings on your screenshots |
| `GitHubRepository.kt` | Error messages (`gh_network_request_failed`, `gh_module_conf_invalid`, etc.) |
| `BuildProgressUtils.kt` | Build progress labels (`bp_build_queued`, `bp_run_waiting_steps`, etc.) |
| `MainViewModel.kt` | Added private `text()` helper → `LocaleHelper.str()` for error paths |

52 new strings added to both `values/strings.xml` (ZH) and `values-en/strings.xml` (EN). Verified parity: **954 ZH keys = 954 EN keys, 0 format placeholder mismatches**.

### 2. Cleanup in `Models.kt`

`BuildProgress.currentStep` default `"等待 GitHub 分配 Runner"` → `""`. The default was never actually visible (UI gates rendering behind `BuildStatus != IDLE` / `totalSteps > 0`, where the field is always assigned via `text()`/`tr()`; `NotificationUtils` already had `isNullOrBlank` guard). Removes the last hardcoded Chinese fallback from a data class default.

xingguangcuican6666:
OK, plz wait for my test. Thanks.

xingguangcuican6666 (approved):

xingguangcuican6666:
seem it good, we will merge it later.

PR #61, opened 2026-05-27
fix(ci,app): drop magiskboot after SukiSU bootimg migration

FunLay123:
---
## Problem
CI fails while bundling ksud (`curl -fsSL`):
```text
curl: (22) The requested URL returned error: 404
https://raw.githubusercontent.com/SukiSU-Ultra/SukiSU-Ultra/main/manager/app/src/main/jniLibs/arm64-v8a/libmagiskboot.so
```
*(Verified: HTTP 404 as of 2026-05-26.)*
ABK also ran a Python patch on `userspace/ksud/src/boot_patch.rs` that replaced an upstream magiskboot copy block with a shell wrapper. That target block was removed upstream together with magiskboot, so the patch would fail with `failed to find magiskboot injection block in boot_patch.rs`.
---
### CI (`.github/workflows/build-abk-app.yml`, `build-abk-app-dev.yml`)
Removed from both workflows:
- Python patch of `boot_patch.rs`
- `magiskboot_url_for_abi()` helper
- `curl` of `libmagiskboot.so` into `userspace/ksud/bin/<abi>/magiskboot` (build-time ksud asset)
Updated build summary line to: *magiskboot removed upstream (ksud uses bootimg crate)*.
Also adds `app_dev/**` to **`build-abk-app.yml`** path filters (upstream `dev` already had this in the dev workflow; release workflow did not).
### App
| File | Change |
|------|--------|
| `RootUtils.kt` | Removed `embeddedMagiskbootPath`, `resolveUserlandMagiskbootPath`, and optional `--magiskboot` args |
| `AbkRootPatchScreen.kt` | Userland boot patch requires bundled `ksud` only (no magiskboot gate) |
| `SettingsScreen.kt` | OSS notice: `android_bootimg` instead of libmagiskboot |
| `THIRD_PARTY_NOTICES.md` | Documents ksud + `android_bootimg`; no libmagiskboot download |
| `app/src/main/res/values/strings.xml`, `values-en`, `values-ru` | Removed `root_patch_warn_no_magiskboot` |
**Diff:** 9 files · **−94 / +9** lines (single commit on branch)
---
## Before → After
| | Before | After |
|---|--------|-------|
| CI ksud build | Downloads missing `libmagiskboot.so` into ksud asset dir | Builds ksud without external magiskboot |
| App `boot-patch` | Could append `--magiskboot <path>` if `libmagiskboot.so` was present | `ksud boot-patch` only (bootimg crate inside ksud) |
| Userland UX | Required ksud **and** magiskboot for non-root patch | Requires ksud only |
> CI has always installed only `libksud.so` into `app/build/generated/abk-jniLibs/main`; magiskboot was a **compile-time ksud asset**, not a separate APK native library.
---

### DIFF
--- a/.github/workflows/build-abk-app-dev.yml
+++ b/.github/workflows/build-abk-app-dev.yml
@@ -73,19 +73,6 @@ jobs:
           commit="$(git -C "$source_dir" rev-parse HEAD)"
           mkdir -p "$target_dir" "$jni_target_dir"
 
-          SOURCE_DIR="$source_dir" python3 - <<'PY'
-          import os
-          from pathlib import Path
-
-          path = Path(os.environ["SOURCE_DIR"]) / "userspace/ksud/src/boot_patch.rs"
-          text = path.read_text()
-          old = """            } else {\n                let magiskboot_path = workdir.join(\"magiskboot\");\n                assets::copy_assets_to_file(\"magiskboot\", &magiskboot_path)\n                    .context(\"copy magiskboot failed\")?;\n                magiskboot_path\n            };\n            ensure!(magiskboot.exists(), \"{} is not exist\", magiskboot.display());\n            #[cfg(unix)]\n            let _ = std::fs::set_permissions(&magiskboot, std::fs::Permissions::from_mode(0o755));\n            magiskboot\n"""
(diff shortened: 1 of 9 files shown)


xingguangcuican6666:
Okay, I found this error now, I will review now.

xingguangcuican6666 (approved):
Seem it good, I will merge it later, thanks!

PR #100, opened 2026-06-09
feat(ui): add Miuix theme style with floating navigation bar

guoyujie666:


xingguangcuican6666 (changes requested):
There cause so many problem and the ref is older than baseline too much.

PR #121, opened 2026-06-25
Fix Workflow step i18n fetch

FunLay123:
httpget func did synced net request on the main thread, but Android prohibits that and says "NetworkOnMainThreadException". The consequence is Workflowstep i18n doesn't work at all. Now it's fixed


xingguangcuican6666:
I will review this pr now, plz wait

xingguangcuican6666 (approved):

PR #146, opened 2026-07-13
feat(miuix): add full MIUIX parallel UI theme

fanziyun:
## Summary

This PR adds a full MIUIX/HyperOS-style parallel UI theme to ABK Manager, sitting alongside the existing Material 3 UI. Users can switch between M3 and MIUIX themes at any time via Settings -> Color & Appearance -> UI Style.

## Device Support (Breaking)

This PR raises `minSdk` from 26 to 33 in both `app/` and `app_dev/`. From this branch onward the app no longer installs or updates on Android 8.0 - 12L devices. **This is a deliberate release decision, not an oversight**, and it supersedes the "full backward compatibility" wording used in earlier revisions of this description.

The floor comes from exactly one dependency: `miuix-blur-android` declares `minSdkVersion 33` in its own manifest. Every other MIUIX artifact (`miuix-ui`, `miuix-core`, `miuix-shader`) only requires 23, so blur is the sole constraint.

Alternatives considered and rejected:

- **Drop blur, keep minSdk 26** -- `layerBackdrop` / `MiuixBlurBackdrop` is used across roughly 25 files. Replacing it with an opaque fallback removes the glass `FloatingBottomBar` and top bars, which are the defining HyperOS visual this PR exists to deliver.
- **Runtime-gate blur behind `Build.VERSION.SDK_INT`** -- the AAR itself declares 33, so this needs a manifest `overrideLibrary` plus a guarantee that no blur class is ever touched on lower API levels. Fragile for the benefit.
- **Split MIUIX into its own flavor/APK** -- doubles build, signing and release surface for one UI mode.

## What's Included

- **Dual-theme architecture**: mutually exclusive M3/MIUIX switching at app entry via `state.uiStyle`, sharing the same `MainViewModel` and data layer. MIUIX screens live under `com.abk.kernel.miuix.*`.
- **20+ MIUIX-styled screens**: Settings, Status, Build, Flash/Files, Module Repository, AuthGate, About, Extension Manager, Root Authorization, SuperUser Profile, Manager Tools, Runtime Home, Installed Modules, Open Source Licenses, Theme Settings, App Profile Templates, Build/Module Repo Settings, plus sub-pages (Plan Library, Build Queue, Module Install Log, Module Action Terminal, custom kernel options).
- **Ported components**: Navigation3 sub-page system (slide animations, predictive back, scrim), FloatingBottomBar (glass-like with blur), TopAppBar with scroll behavior, animated SearchBar, MIUIX WindowDialogs, theme-aware SnackbarHost, and HyperOS-styled hero cards/cards/chips/switches/dropdowns.
- **Visual effects**: dynamic blur, Liquid Glass (lens, vibrancy, inner shadow), Monet theme support (key color, palette style, color spec), scroll-end haptics.
- **Build screen**: all 13 M3 sections ported (hero card, plan tools, target selector, progress, kernel version, KernelSU, features, ZRAM/KPM, custom modules, config, submit, recent runs with expand/collapse).
- **i18n**: all MIUIX-specific strings localized in `values/`, `values-en/`, `values-ru/`.
- **Technical**: MIUIX v0.9.2, Kotlin 2.4.0, minSdk 33 (see Device Support above), legacy DEX packaging, Navigation3 with parcelable routes, reusable terminal state/params helpers, root shell caching.
- **Testing**: new unit tests for `AuthOobeCoordinator`, `FlashTerminalLogState`, `FlashTerminalParams`. `:app:compileDebugKotlin` passes with no new warnings. Note that `:app:testDebugUnitTest` cannot currently execute on this branch **or on `dev`**: `app/src/test/java/com/abk/kernel/data/repository/GitHubRepositoryNetworkTest.kt` no longer compiles against the current `GitHubRepository` / `Result` API (`getRepo` does not exist). That file is byte-identical to `dev`, so the breakage is pre-existing and not introduced here.

## Compatibility with Existing Material 3 UI

- **M3 behaviour is preserved** -- switching `uiStyle` back to `"material"` restores the original experience. This is UI-level compatibility only; see Device Support above for the install-level break.
- **Single source of truth** -- both themes share the same `MainViewModel`, data layer, and string resources. No business logic is forked per theme.
- **Business logic once, UI twice** -- add logic to `MainViewModel`, then implement screens in both `ui/screens/` (M3) and `miuix/ui/screens/` (MIUIX).
- Theme color settings are independent per mode -- switching themes does not carry over custom accent colors.
- Navigation UX differs: M3 uses boolean-based child pages with `AnimatedContent`; MIUIX uses a Navigation3 back stack with slide animations and predictive back gestures.
- **Navigation bar differs from upstream** -- upstream (M3) uses the standard `NavigationBar`/`NavigationRail` with Compose's `AnimatedContent` tab switching. MIUIX replaces this with a custom `FloatingBottomBar` (glass-like floating nav with blur) backed by `Navigation3` + `HorizontalPager`, which changes the tab switching mechanism, animation behavior, and sub-page navigation model entirely.

### Shared and M3 files this PR does touch

MIUIX screens are isolated under `com.abk.kernel.miuix.*`, but the branch is **not** confined to that package. For reviewer clarity, everything outside it:

| Path | Reason |
| --- | --- |
| `MainActivity.kt` | theme branching point and MIUIX OOBE overlay |
| `ui/navigation3/{Navigator,Routes}.kt` | new shared Navigation3 back stack (new files) |
| `ui/component/liquid/{Lens,InnerShadow,Vibrancy,CombinedBackdrop}.kt` | Liquid Glass primitives, shared (new files) |
| `viewmodel/MainViewModel.kt`, `viewmodel/AuthOobeCoordinator.kt` | `uiStyle` state and the OOBE entry both themes call |
| `data/repository/PreferencesRepository.kt` | persists `uiStyle` and the MIUIX color keys |
| `extensions/AbkExtensionHost.kt`, `extensions/AbkExtensionManagerActivity.kt` | extension host lifted out of the activity so both themes can drive it |
| `ui/screens/AuthGateScreen.kt` | OOBE theme-selection step |
| `ui/screens/SettingsScreen.kt` | UI Style picker, license notices |
| `ui/screens/RuntimeScreens.kt`, `ui/screens/ModuleRepositoryScreen.kt`, `ui/screens/BuildScreen.kt` | UI-style plumbing plus the shared fixes listed below |
| `res/values*/strings.xml` | MIUIX chrome strings, shared across both themes |

## Review Fixes Applied

Addressing review feedback on earlier revisions:

- **ZRAM custom algorithm field** -- rewriting the M3 branch as `AnimatedVisibility` had dropped the negation, so the field appeared only when "full algorithm support" was *enabled*, hiding it in the configuration where `zramExtraAlgos` actually applies. Restored to `!config.zramFullAlgo` in both themes.
- **Module repository loading state** -- keeping previous items visible during a recompute also cleared the `loading` flag before the merge/filter work started, so `computing` reported a stale list as fully up to date. The flag is honest again; the full-bleed placeholder now only takes over when there is nothing stale left to show, and an ongoing recompute over existing rows surfaces as an inline indicator instead. Applied to the runtime and build lists in both themes.
- **MIUIX Monet** -- `keyColor` fell back to a hardcoded seed and was therefore never `null`, so miuix could never reach `platformDynamicColors()`. The "Monet" toggle generated from that seed instead of the wallpaper, contradicting its own description and diverging from M3. Default now yields the real system palette; a user-picked accent color still seeds the scheme.
- **Settings account row** -- the MIUIX "not signed in" row is now a login entry into the OOBE flow, matching the M3 row from #178.
- **Dependency notices** -- `THIRD_PARTY_NOTICES.md` and both in-app license screens now agree with `gradle/libs.versions.toml`: Kotlin corrected to 2.4.0 in all three places, and the M3 screen no longer omits `navigation3-runtime`, `lifecycle-viewmodel-navigation3` and `lifecycle-process`.
- **Stray UTF-8 BOM** -- removed from every touched source. `viewmodel/RuntimeCoordinator.kt` had no change other than the BOM and is byte-identical to `dev` again.

## AI Disclosure

Portions of the code in this PR, including the MIUIX screen implementations, component ports, and translations, were AI-assisted and have passed local testing.


xingguangcuican6666 (approved):

app-screenshot.png
xingguangcuican6666, line 1: Is it necessary?

xingguangcuican6666 (changes requested):

app-screenshot.png
fanziyun: Oh okay I revert it

app/build.gradle.kts
xingguangcuican6666, line 1: https://github.com/xingguangcuican6666/ABK/blob/miuix/app_dev%2Fbuild.gradle.kts#L13-L15
there version code is unchanged

xingguangcuican6666 (changes requested):

fanziyun:
oh This PR is still unstable

xingguangcuicanbot (changes requested):
## Automated PR Review

- Repo: xingguangcuican6666/ABK
- PR: #146
- Request user: xingguangcuican6666
- Diff lines: 30069
- Final action: request_changes
- Codex recommendation: request_changes

### Summary
Two regressions: a hard `minSdk` jump to 33 and a dead onboarding theme-selection path.

### Findings
- [high] app/build.gradle.kts: `minSdk` was raised from 26 to 33, which makes this branch uninstallable on Android 8-12 devices and is a major compatibility regression unrelated to the UI theme work. Suggestion: Keep `minSdk` at the previous supported level unless the product intentionally drops older Android versions and the release notes reflect that decision.
- [medium] app/src/main/java/com/abk/kernel/viewmodel/AuthOobeCoordinator.kt: The new `AuthStep.THEME_SELECT` state is effectively unreachable: `completeIfRequested(true)` is only called from the build/fork flow, but that path always sets `oobeFromBuild = true`, so the `!state.oobeFromBuild` branch never runs. Suggestion: Add an actual caller for the theme-selection step in the initial onboarding flow, or remove the dead state/branch if onboarding theme selection is not intended.

PR URL: https://github.com/xingguangcuican6666/ABK/pull/146

app/build.gradle.kts
```diff
@@ -76,7 +78,7 @@ android {
 
     defaultConfig {
         applicationId = "com.abk.kernel"
-        minSdk = 26
```
xingguangcuicanbot, line 79: 位置: app/build.gradle.kts:81

问题: `minSdk` was raised from 26 to 33, which drops installation/updates on Android 8-12 devices and contradicts the backward-compatibility claim in the PR description. If this is meant to be a theme add-on, it should not force the whole app onto API 33.

建议: Keep the app minSdk unchanged, or isolate the MIUIX-only code behind an API-compatible implementation.

app/src/main/java/com/abk/kernel/ui/screens/BuildScreen.kt
```diff
@@ -1489,7 +1489,7 @@ fun BuildScreen(
                     SwitchRow(stringResource(R.string.build_zram_full_algo), config.zramFullAlgo) {
                         vm.updateBuildConfig(config.copy(zramFullAlgo = it))
                     }
-                    if (!config.zramFullAlgo) {
+                    AnimatedVisibility(config.zramFullAlgo) {
```
xingguangcuicanbot: 位置: app/src/main/java/com/abk/kernel/ui/screens/BuildScreen.kt:1492

问题: This visibility condition is inverted relative to the existing setting label: the custom ZRAM algorithm field now appears only when `zramFullAlgo` is enabled, so the user can no longer edit `zramExtraAlgos` in the mode where it is actually used.

建议: Restore the previous condition so the extra algorithm field is shown when `zramFullAlgo` is false.

app/src/main/java/com/abk/kernel/miuix/ui/screens/ThemeSettingsScreenMiuix.kt
```diff
@@ -0,0 +1,295 @@
...
+                            tint = MiuixTheme.colorScheme.onBackground
+                        )
+                    },
+                    checked = dynamicColorAvailable && miuixState.miuixDynamicColorEnabled,
+                    onCheckedChange = { enabled ->
+                        miuixVm.setMiuixDynamicColorEnabled(enabled)
+                    }
+                )
+
+                AnimatedVisibility(
```
xingguangcuicanbot, line 135: 位置: app/src/main/java/com/abk/kernel/miuix/ui/screens/ThemeSettingsScreenMiuix.kt:135

问题: The MIUIX key color / palette controls are only rendered while Monet is enabled, but `AbkMiuixTheme` still uses `miuixThemeColorArgb`, `miuixColorStyle`, and `miuixColorSpec` when Monet is off. Turning Monet off therefore hides the only controls for configuring the static MIUIX palette.

建议: Show these controls when Monet is disabled, or otherwise provide a separate configuration path for the non-dynamic theme path.

xingguangcuicanbot (changes requested):
## 自动 PR Review

### 总结
Three regressions: a global minSdk jump, an inverted ZRAM field condition, and MIUIX theme controls hidden in the wrong mode.

### 结论
request_changes

### 说明
Codex 建议: request_changes
已留下 3 条行级评论。

xingguangcuican6666:
Don't notice there of requested changes, that just for test my auto review agent.

app/build.gradle.kts
xingguangcuicanbot, line 79: Location: app/build.gradle.kts:81

Issue: Bumping `minSdk` to 33 drops install support for every Android 12L-and-below device. That is a hard compatibility regression for a theme/UI PR and will prevent existing users on those devices from upgrading or installing the app.

Suggestion: Keep the previous `minSdk` unless you are intentionally cutting off older Android versions, and call that out explicitly if so.

app_dev/build.gradle.kts
```diff
@@ -76,10 +78,10 @@ android {
 
     defaultConfig {
         applicationId = "com.abk.kernel"
-        minSdk = 26
+        minSdk = 33
         targetSdk = 35
-        versionCode = 10025
-        versionName = "1.2.5-dev"
```
xingguangcuicanbot: Location: app_dev/build.gradle.kts:83

Issue: This dev build is now packaged as `versionCode = 10023` / `versionName = 1.2.3-dev`, while the rest of the build still reports the higher 10025/1.2.5-dev values used by update checks. That mismatch will confuse in-app version reporting and can also block in-place upgrades from existing dev installs because Android sees this as a downgrade.

Suggestion: Restore the dev `versionCode`/`versionName` to match the build constants, or bump them above the previous dev release if this is meant to be installable over it.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
Two release-blocking versioning regressions: the app now requires Android 13+, and the dev flavor was downgraded to a lower versionCode/versionName than the build still reports internally.

### Decision
request_changes

### Notes

app/build.gradle.kts
xingguangcuicanbot, line 79: Location: app/build.gradle.kts:81

Issue: Raising `minSdk` from 26 to 33 turns an optional UI theme into an app-wide compatibility break: the APK built from this branch will no longer install or update on Android 8-12L devices. Nothing else in the PR scopes this to the MIUIX path, so all existing users on API 26-32 are dropped as soon as this ships.

Suggestion: Keep the current minimum SDK unless dropping Android 8-12L support is an explicit release decision with a migration plan. If MIUIX really needs API 33+, gate that theme or those features behind runtime/API checks instead of raising the whole app's floor.

app_dev/build.gradle.kts
xingguangcuicanbot: Location: app_dev/build.gradle.kts:83

Issue: The dev variant version was rolled back from `1.2.5-dev` / `10025` to `1.2.3-dev` / `10023`. That makes freshly built dev APKs look older than already-published builds, which breaks normal upgrade paths and leaves the app-update metadata inconsistent with the actual package version.

Suggestion: Restore the dev variant version to at least the current published value so installs and update checks remain monotonic.

```suggestion
versionCode = 10025
versionName = "1.2.5-dev"
```

app/src/main/java/com/abk/kernel/miuix/ui/screens/BuildScreenMiuix.kt
```diff
@@ -0,0 +1,3628 @@
...
+            Column(verticalArrangement = Arrangement.spacedBy(12.dp)) {
+                SectionTitle(stringResource(R.string.build_zram_options))
+                Card(modifier = Modifier.fillMaxWidth()) {
+                    SwitchPreference(
+                        title = stringResource(R.string.build_zram_full_algo),
+                        checked = config.zramFullAlgo,
+                        onCheckedChange = { vm.updateBuildConfig(config.copy(zramFullAlgo = it)) }
+                    )
+                    AnimatedVisibility(
+                        visible = config.zramFullAlgo,
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/miuix/ui/screens/BuildScreenMiuix.kt:1656

Issue: This makes the "Custom ZRAM algorithms" field appear only when `zramFullAlgo` is enabled. In the existing UI, that field was shown for the opposite state, so the new MIUIX build screen now hides custom algorithm input in the configuration where users previously entered it. That is a behavioral regression and can silently prevent users from editing `zramExtraAlgos`.

Suggestion: Match the previous visibility condition so custom ZRAM algorithm input remains available in the same configuration as before.

```suggestion
visible = !config.zramFullAlgo,
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
I found regressions in platform compatibility, dev build versioning, and the ZRAM options UI that should be addressed before merging.

### Decision
request_changes

### Notes

guoyujie666:
后继有人啊

fanziyun:
emm This PR is temporarily on hold, but I will continue to keep it in sync with upstream

app/build.gradle.kts
xingguangcuicanbot, line 79: Location: app/build.gradle.kts:81

Issue: Raising the app's `minSdk` from 26 to 33 is a breaking compatibility regression for every existing user on Android 8-12L. This PR is presented as a theme addition, but this line makes the release uninstallable on a large part of the current install base unless you split the feature by flavor or find a lower-API implementation. The same change is mirrored in `app_dev/build.gradle.kts` as well.

Suggestion: Restore the previous minimum SDK, or isolate the MIUIX-only implementation behind a separate flavor/build so the existing app variant keeps its current device support.

app/src/main/java/com/abk/kernel/ui/screens/ModuleRepositoryScreen.kt
```diff
@@ -183,7 +183,9 @@ fun ModuleRepositoryScreen(
             value = ModuleListComputation(items = emptyList(), loading = false)
             return@produceState
         }
-        value = ModuleListComputation(loading = true)
+        // Keep previous items visible while recomputing to avoid UI flicker
+        val previousItems = value.items
+        value = ModuleListComputation(items = previousItems, loading = false)
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/ui/screens/ModuleRepositoryScreen.kt:188

Issue: This change keeps the old repository items visible during recomputation, but it also flips `loading` to `false` before the expensive merge/filter work starts. Downstream code uses that flag for `computing`, so the screen now treats stale results as fully up to date: progress indicators stop, empty/loading states cannot distinguish refresh-in-progress, and users can interact with modules that may already have disappeared from the refreshed repository set. The same regression is repeated in the build-repository branch later in this file.

Suggestion: Preserve the previous items, but keep `loading = true` until the recomputation finishes so the UI can show that the visible list is stale.

```suggestion
val previousItems = value.items
value = ModuleListComputation(items = previousItems, loading = true)
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The PR introduces one release-blocking platform regression and two behavioral regressions in onboarding/repository-state handling.

### Decision
request_changes

### Notes

- app/src/main/java/com/abk/kernel/ui/screens/AuthGateScreen.kt:108 The initial onboarding flow now sends already-signed-in users through `openBuildOobe()`. That method marks the flow as `oobeFromBuild = true`, and `completeIfRequested()` skips the new `THEME_SELECT` step whenever `oobeFromBuild` is true. As a result, authenticated users who hit the first-run OOBE never see the UI-style picker and complete onboarding without the theme choice this PR adds.

app/build.gradle.kts
fanziyun: This PR raises `minSdk` from 26 to 33 in both `app/` and `app_dev/`.
From this branch onward the app no longer installs or updates on
Android 8.0 - 12L devices. This is a deliberate release decision,
not an oversight.

xingguangcuicanbot, line 79: Location: app/build.gradle.kts:81

Issue: This raises the app's install floor from Android 8.0 (API 26) to Android 13 (API 33). Nothing else in the PR updates the product contract or provides a compatibility fallback, so merging this will silently strand every existing user on API 26-32 from receiving upgrades. That is a major regression for the currently shipped app.

Suggestion: Restore the previous `minSdk` unless dropping support for Android 8-12 is an intentional product decision that is coordinated separately.

```suggestion
minSdk = 26
```

app_dev/build.gradle.kts
```diff
@@ -76,7 +78,7 @@ android {
 
     defaultConfig {
         applicationId = "com.abk.kernel"
-        minSdk = 26
```
xingguangcuicanbot, line 79: Location: app_dev/build.gradle.kts:81

Issue: The dev build keeps the same unintended platform drop as the release build. If this branch is merged as-is, nightly/dev APKs will also stop installing on API 26-32 devices, which blocks validation on the same devices the stable app still supports today.

Suggestion: Keep the dev variant aligned with the supported device floor, or explicitly document and coordinate a platform-support break before landing this change.

```suggestion
minSdk = 26
```

app/src/main/java/com/abk/kernel/ui/screens/SettingsScreen.kt
```diff
@@ -2339,6 +2396,7 @@ private fun openSourceNoticeGroups(): List<OpenSourceNoticeGroup> = listOf(
```
xingguangcuicanbot, line 2346: Location: app/src/main/java/com/abk/kernel/ui/screens/SettingsScreen.kt:2401

Issue: The dependency notice shown inside the app still claims the Kotlin Gradle/Compose plugin is `2.3.21`, but this PR upgrades the build to Kotlin `2.4.0` in `gradle/libs.versions.toml`. Shipping stale OSS attribution makes the in-app license screen inaccurate.

Suggestion: Update the in-app dependency notice so it matches the actual Kotlin version declared by this PR.

```suggestion
OpenSourceNotice("Kotlin Gradle/Compose plugin 2.4.0", "Apache-2.0", "org.jetbrains.kotlin.plugin.compose"),
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The PR introduces a major backward-compatibility regression by raising both app variants to `minSdk 33`, and it leaves the OSS/license notices out of sync with the Kotlin 2.4.0 upgrade.

### Decision
request_changes

### Notes

- THIRD_PARTY_NOTICES.md:63 `THIRD_PARTY_NOTICES.md` was updated for the new dependencies, but the Kotlin plugin entry still says `2.3.21` even though this PR bumps Kotlin to `2.4.0`. That leaves the bundled notices inconsistent with the shipped build metadata.

app/src/main/java/com/abk/kernel/viewmodel/AuthOobeCoordinator.kt
```diff
@@ -29,18 +29,25 @@ class AuthOobeCoordinator(
...
                 showOobe = true,
+                oobeFromBuild = false,
                 authStep = AuthStep.INTRO,
                 error = null,
             )
         }
     }
 
-    fun openBuildOobe() {
+    fun openBuildOobe() = enterOobeFlow()
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/viewmodel/AuthOobeCoordinator.kt:39

Issue: `openBuildOobe()` now reuses `enterOobeFlow()`, which unconditionally sets `oobeFromBuild = true`. On a first launch where the user is already logged in, tapping Continue from the intro goes through `openBuildOobe()`, so `completeIfRequested(true)` treats that path as a build-only re-entry and skips the new `THEME_SELECT` step entirely. That means one class of first-run users can never choose the UI style during OOBE.

Suggestion: Keep the existing settings/login re-entry behavior, but let the initial build/onboarding path preserve `oobeFromBuild = false` so first-run users still reach `THEME_SELECT` after auth/fork checks.

```suggestion
fun openBuildOobe() = enterOobeFlow(oobeFromBuild = false)

/** Settings account entry: jump back into the OOBE flow so the user can sign in again. */
fun openLoginOobe() = enterOobeFlow(oobeFromBuild = true)

private fun enterOobeFlow(oobeFromBuild: Boolean) {
    val state = readState()
    val nextStep = if (state.isLoggedIn && state.user != null) AuthStep.FORK_CHECK else AuthStep.LOGIN
    updateState {
        it.copy(
            showOobe = true,
            oobeFromBuild = oobeFromBuild,
            authStep = nextStep,
            error = null,
        )
    }
    if (state.isLoggedIn && state.user == null) {
        scope.launch {
            val user = fetchUser(true) ?: return@launch
            updateState { it.copy(authStep = AuthStep.FORK_CHECK, user = user, isLoggedIn = true) }
            requestForkCheck(true)
        }
    } else if (nextStep == AuthStep.FORK_CHECK) {
        requestForkCheck(true)
    }
}
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
I found one functional regression: the new shared OOBE entry path skips the theme-selection step for first-launch users who are already signed in.

### Decision
request_changes

### Notes

PR #147, opened 2026-07-16
fix(cli): stabilize machine API and preserve Android/Desktop compatibility

Mazha0309:
## Summary

This PR makes the ABK CLI a stable dependency for Desktop automation while keeping the Android app and existing build workflows authoritative. It changes only CLI code, catalogs, documentation, tests, dependencies, and the CLI packaging workflow; Android, Desktop, and other CI implementation files are not modified.

## Root cause

Desktop invokes the CLI as a machine-readable backend, but the merged CLI did not yet provide the required global JSON contract or targeted artifact operations. The review also found mismatches with Android/workflow inputs, repository-scoped credentials, frozen runtime packaging, locale compatibility, redirect handling, and concurrent artifact pagination.

## What changed

- Add a versioned `--json` contract that emits exactly one UTF-8 JSON document with structured errors and normalized results for authentication, repository, build, status, artifact, list, and self-test operations.
- Keep JSON mode non-interactive and add `self-test` coverage for frozen runtime dependencies.
- Preserve explicit `--repo` / `ABK_REPO` targets instead of silently switching to a user fork. OAuth/PAT credentials retain user identity, while GitHub App installation tokens can operate without the unsupported `/user` requirement.
- Align CLI build plans and validation with the Android/workflow contracts, including `Latest`, KPM/SUSFS/virtualization normalization, OnePlus inputs, custom-LTS pairs, revision support, and unsafe free-form input rejection.
- Harden secret redaction, repository-scoped signing state, cross-process atomic config updates, Android-compatible signing-key reconciliation, HTTPS handling, bounded downloads, release URL validation, truncation checks, and atomic artifact replacement.
- Follow authenticated GitHub API redirects only when they remain HTTPS and same-origin; cross-origin downloads continue to strip credentials.
- Add artifact selection by ID and multi-page listing that tolerates shifted pages without duplicates or omissions.
- Return real workflow-run state after dispatch when available; if GitHub has not exposed the run yet, keep dispatch metadata without inventing a `queued` state.
- Expose canonical BCP 47 language tags and private-use tags while accepting legacy IDs and locale variants. Shared config continues to store legacy-compatible IDs for older CLI/Desktop readers.
- Bundle the required frozen dependencies, copy the repository license locally, and expand CLI-only packaging smoke tests and regression coverage.

## Compatibility

- Existing lowercase language values and `jp-neko`, `zh-neko`, and `zh-zako` remain accepted.
- Older CLI/Desktop builds can still read the shared language configuration.
- Existing Android workflows are consumed as contracts but are not changed by this PR.
- Normal Desktop OAuth/PAT responses retain the existing `user` object; only repository-scoped installation-token sessions return `user: null`.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 python3 -m unittest discover -s cli/tests -v` — 136 tests passed
- `python3 cli/abk.py --json self-test`
- `git diff --check`
- The CLI packaging workflow runs regression and frozen-binary smoke tests across its Windows and Linux target matrix.


cli/abk.py
```diff
@@ -2474,15 +3883,32 @@ def main():
...
         return 0
 
+    if _target_repo(args) and args.command in {"login", "sync"}:
+        source = "--repo" if args.repo else "ABK_REPO"
+        parser._command_hint = args.command
+        parser.error(f"{source} is not supported by {args.command}")
+
+    if args.json:
+        return _run_json_command(args)
+    _persist_requested_language(args.lang)
```
Mazha0309, line 4059: [P2] Persist `--lang` before `parse_args()` can exit for `--help`

Language persistence now happens after argument parsing, but the argparse help action raises `SystemExit` during `parse_args()`. Consequently, the documented command `abk --lang en-us --help` displays English but does not write the selected language to the config. I reproduced this with a temporary `XDG_CONFIG_HOME`: exit code 0 and no config file. `upstream/dev` persisted the valid early language before constructing the parser, so this is a regression introduced by this PR.

Please persist a valid `early_language` before the help action can exit, or make the help action perform the persistence explicitly.

cli/abk.py
```diff
@@ -589,31 +712,40 @@ def __init__(self, token=None, repo=None, verbose=False):
...
+        self.repo_explicit = bool(self.repo)
         self.verbose = verbose
         self.username = None
         self.fork_repo = None
+        self.authentication_error = None
+        self.fork_detection_error = None
         
-        if self.token:
-            self._detect_user(detect_fork=not self.repo_explicit)
+        if self.token and not self.repo_explicit:
```
Mazha0309, line 784: [P2] Avoid requiring a user identity for explicit-repository `whoami`

This branch intentionally skips the `/user` probe for an explicit `--repo`, allowing repository-scoped credentials such as GitHub App installation tokens to operate directly on the target repository. However, `cmd_whoami()` still calls `client.get_user()` unconditionally. The GitHub `GET /user` endpoint does not support installation tokens, while workflow dispatch does, so a valid token may be able to trigger a build but still fail `abk --json --repo org/ABK whoami`.

For explicit-repository mode, please validate repository access and return `user: null` without calling `/user`. A real 401 should also map to the existing `not_authenticated` code instead of always becoming `login_verification_failed`.

Docs: https://docs.github.com/en/rest/users/users#get-the-authenticated-user and https://docs.github.com/en/rest/actions/workflows#create-a-workflow-dispatch-event

Mazha0309 (commented):
The review confirmed two reproducible CLI issues, both annotated inline. The comments cover only regressions introduced by this PR or newly supported paths that remain incomplete; they do not include Android, Desktop, other CI, or pre-existing upstream issues.

Mazha0309:
Ready for maintainer review.

Validation for head `ab29b283` is complete:

- the full **Build ABK CLI** matrix passed on Windows, macOS, and Linux, including x86, x64, ARM64, and ARMHF targets;
- all frozen-binary smoke tests passed;
- 136 local CLI regression tests passed;
- `abk --json self-test` and `git diff --check` passed;
- the reported CLI review findings and final audit follow-ups have been addressed.

The changes remain limited to the CLI and its own packaging workflow. Android, Desktop, and other CI implementations are unchanged.

Branch protection still requires an approving review from another reviewer with write access; GitHub does not count approval from the PR author.

cli/abk.py
```diff
@@ -2001,9 +3001,24 @@ def cmd_build(args):
...
+                            f"({resolve_plan_ksu_branch(variant, args.ksu_branch)})"
+                        )
+                        warning = t("op_no_kpm_ksu", ksu=selection)
+                        warning_messages.append(warning)
+                        print(t("warning_prefix") + " " + warning)
+                    kernel_version = (
+                        args.kernel_version or "5.10"
+                        if target == "custom"
+                        else workflow["kernel"]
+                    )
```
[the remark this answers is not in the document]
Mazha0309: Thanks for flagging this. This is not reproducible because Python parses the expression as (args.kernel_version or "5.10") if target == "custom" else workflow["kernel"]; the conditional expression has lower precedence than or. I verified the suggested case with abk --json build --matrix a14 --kernel-version 6.12 --virt 678 --dry-run: it dispatches kernel-a14-6-1.yml with virtualization_support set to "678", confirming that fixed targets use the workflow kernel version. The related workflow-contract test also passes, so no code change is needed here.

xingguangcuican6666 (approved):

PR #148, opened 2026-07-16
feat(signing): manage CLI keys with Android rotation compatibility

Mazha0309:
## Summary

- add `abk signing status|import|rotate|enable|disable` for fork-scoped artifact signing key management
- validate imported RSA SPKI/PKCS#8 key pairs, keep private material out of local config and command output, and publish only the verification key through the existing release asset
- protect key installation and disablement with fresh remote snapshots, exact release-asset IDs, concurrent-pair preservation, and a repo-scoped safety lock for indeterminate states
- keep login, fork, and sync successful when only signing repair is blocked, while CLI build and artifact verification remain fail closed
- make signing `--dry-run --json` fields describe the projected post-operation state consistently
- introduce CLI version `0.1.0` via `abk --version`, `abk --json --version`, and the additive `cliVersion` JSON field
- minimally update Android compatibility: validate remote RSA/SPKI keys, refuse invalid or partial signing state without regenerating it, and refresh a stable public key immediately before verifying trusted Kernel/AnyKernel bundles

No Desktop code or GitHub Actions workflow was changed. Existing CLI artifact names remain unchanged.

## Why

The CLI could initialize signing material but could not import user-managed keys, rotate an existing pair, or intentionally disable artifact verification. GitHub Actions Secrets and release assets are also not atomic, so stale status reads, concurrent Android updates, or lost API responses could otherwise delete or mismatch signing material.

This change adds explicit management commands and conservative transaction checks. A mutation is authorized only against the remote key/Secret state that the command inspected. If a post-write state cannot be proven, the CLI records an indeterminate safety lock that blocks later CLI builds/downloads until the user repairs or disables signing.

Android remains the primary client. Its changes are intentionally limited to CLI interoperability: malformed remote keys cannot replace a valid cache, partial state cannot trigger automatic key regeneration, and trusted bundle verification refreshes the remote key with a public–Secret–public stability check. Untrusted APK/module/legacy downloads do not invoke this refresh.

## Validation

- `python3 -m unittest discover -s cli/tests -p 'test_*.py'` — 194 passed
- `python3 -m compileall -q cli` — passed
- `ANDROID_HOME=/home/mazha0309/android-sdk ANDROID_SDK_ROOT=/home/mazha0309/android-sdk gradle :app:compileDebugKotlin` — passed
- `git diff --check` — passed

## Existing test-suite issue

`:app:testDebugUnitTest` does not reach test execution because the existing `RuntimeCoordinatorStatePatchTest.kt` source currently fails to compile with an unresolved `MainUiState` reference. The changed Android main source compiles successfully; this PR does not modify that pre-existing test.


cli/abk.py
```diff
@@ -2479,6 +3182,368 @@ def cmd_sync(args):
...
+        args,
+        action=args.signing_action,
+    ):
+        return 1
+
+    try:
+        repo = _select_signing_repository(client, args)
+        args._signing_repo = repo
+
+        status = get_signing_status(client)
```
Mazha0309, line 3401: **[P1] Revalidate this snapshot inside the install transaction.** Confirmation and branch selection rely on this early read. If it reports `absent` and Android completes pair B before `install_signing_keypair`, `rotate`/`import` require no confirmation and `enable` takes its generate path; the installer then treats B as old material and deletes/replaces it. I reproduced the destructive `secret_delete -> public_replace -> secret` sequence. Pass the expected remote state/key/Secret presence into the locked transaction and abort or adopt when it changed.

Mazha0309 (commented):
I reviewed the current PR head, including the CLI 0.1.0 version-reporting follow-up. The version addition is clean and the full 181-test CLI suite passes. I found six reproducible issues in the new signing-management and Android compatibility paths; details are inline. This is a COMMENT review only.

cli/abk.py
```diff
@@ -1673,7 +2034,345 @@ def _ensure_signing_key_locked(client):
...
+        except Exception as exc:
+            raise SigningStateIndeterminateError(
+                "GitHub did not confirm whether the signing public key was deleted"
+            ) from exc
+        if public_key_still_exists:
+            raise SigningStateIndeterminateError(
+                "a signing public key appeared while verification was being "
+                "disabled; the concurrent asset was not touched"
+            )
+        _remove_reappeared_signing_secret(
```
Mazha0309: **[P1] Preserve a concurrently completed pair during the final disable check.** After public-key absence is confirmed above, Android can publish `public B + Secret B` before this call. This helper sees only that a Secret exists and deletes B's Secret; disable then saves a normal disabled state without rechecking the public asset. The result is public B with no Secret and no indeterminate lock. Recheck both resources before cleanup, touch neither when a public key appeared, and persist the indeterminate state.

cli/abk.py
```diff
@@ -1272,6 +1302,140 @@ def publish_signing_key(self, public_key_pem):
...
+        release = self.get_release_by_tag(SIGNING_RELEASE_TAG)
+        if not release:
+            return None
+        signing_assets = [
+            asset for asset in self.list_release_assets(release["id"])
+            if asset.get("name") == SIGNING_PUBLIC_KEY_ASSET
+        ]
+        previous_key = None
+        if signing_assets:
+            previous_key = self._download_release_asset_text(signing_assets[0]["url"])
```
Mazha0309: **[P2] Avoid this second asset download after the Secret has been deleted.** `disable_signing_verification` already captured `old_public_key`. If this redundant read has a transient failure, public deletion never starts, leaving a public-only remote state; because the exception is not `SigningStateIndeterminateError`, no safety lock is persisted. Delete the snapshotted asset IDs without downloading again, or convert this post-Secret failure into an indeterminate result.

app/src/main/java/com/abk/kernel/viewmodel/MainViewModel.kt
```diff
@@ -1458,6 +1454,10 @@ class MainViewModel @JvmOverloads constructor(
                     }
                 }
             }
+            if (secretExists && !existingPublicKey.isNullOrBlank()) {
```
Mazha0309: **[P2] Validate and refresh the remote key at the point it is consumed.** The new remote-first branch accepts any nonblank body and overwrites a valid cached key without parsing it as an RSA/SPKI key. It also runs only from `checkFork`/`forkRepo`, so an App kept open while the CLI rotates A to B still verifies the next download with cached A until another fork check or restart. Validate with `ForkSigningManager` before saving, retain the valid fallback on malformed assets, and refresh before artifact verification.

cli/abk.py
```diff
@@ -2479,6 +3182,368 @@ def cmd_sync(args):
...
+            if args.dry_run:
+                _set_json_result(
+                    args,
+                    ok=True,
+                    action=action,
+                    repo=repo,
+                    dryRun=True,
+                    changed=changed,
+                    verificationEnabled=True,
+                    signingKeyConfigured=status["signing_key_configured"],
```
Mazha0309: **[P2] Make dry-run state fields consistently current or projected.** Importing into an empty repository reports `verificationEnabled: true` but `signingKeyConfigured: false` and `signingReady: false`; successful non-dry-run responses use the same fields for final state. Enable and disable have the same mixed semantics, so automation cannot interpret them reliably. Return projected post-operation values, or expose explicitly named current/planned state objects.

cli/abk.py
```diff
@@ -1538,10 +1890,19 @@ def _create_or_update_signing_secret(
     ) from last_error
 
 
-def _ensure_signing_key_locked(client):
+def _ensure_signing_key_locked(client, *, force_enable=False):
     config = load_config()
     external_key = os.environ.get("ABK_SIGNING_KEY")
     state = _get_signing_state(config, client.repo)
+    if state.get("indeterminate") is True:
```
Mazha0309, line 1912: **[P2] Do not turn unrelated successful operations into signing-lock failures.** `login`, `fork`, and `sync` all call `ensure_signing_key`; for a behind fork, `sync_fork` can complete and print success before this new exception is raised, after which the command returns `sync_failed`. The README says this lock blocks build/download. Either keep signing setup out of those operations when locked or report their success with a separate signing warning/error.

xingguangcuican6666 (approved):

PR #149, opened 2026-07-16
feat(app): module updates

fanziyun:
Use each installed module's updateJson and metamodule state from ksud so update discovery, public Downloads installs, changelog display, META ordering, and pending-reboot behavior match the installed-module runtime flow.


fanziyun:
wait let me resolve this first

fanziyun:
- Add `downloadRuntimeModuleAsset()` helper in `DownloadUtils` to download module zips into `Downloads/ABK` with `preserveDownloadedZip = true`
- Update `ModuleRepositoryScreen` runtime module installation to use `downloadRuntimeModuleAsset()`
- Add public `fileSha256Hex()` utility to `DownloadUtils` for SHA-256 checksum computation
- Replace private duplicate in `RuntimeScreens` with the shared `DownloadUtils.fileSha256Hex()`
- Support both `"sha256"` and `"sha_256"` field names when parsing update JSON

- Local compilation passes (`./gradlew :app:compileDebugKotlin` → BUILD SUCCESSFUL)
- All 103 unit tests pass, no new failures introduced

fanziyun:
## Security Verification Flow: Download to Install

The runtime module update and installation flow enforces a multi-layered security chain from URL validation through checksum verification before flashing.

---

### 1. HTTPS-Only URL Validation

All external URLs (update JSON, zip download, changelog) are required to use HTTPS. Non-HTTPS URLs are rejected or returned as raw text without fetching.

```kotlin
internal fun isSecureRuntimeModuleUrl(url: String): Boolean =
    url.trim().startsWith("https://", ignoreCase = true)
```

- `fetchRuntimeModuleUpdateInfo()` — returns `null` if the update JSON URL is not HTTPS
- `resolveRuntimeModuleChangelog()` — returns the raw URL string for non-HTTPS (safe fallback), only fetches HTTPS URLs
- The `zipUrl` field from update JSON is also validated: if not HTTPS, the entire update is rejected

```kotlin
internal suspend fun fetchRuntimeModuleUpdateInfo(updateJson: String): RuntimeModuleUpdateInfo? {
    if (!isSecureRuntimeModuleUrl(updateJson)) return null
    val result = fetchRuntimeModuleResponse(updateJson).orEmpty()
    if (result.isBlank()) return null
    val json = runCatching { JsonParser.parseString(result).asJsonObject }.getOrNull() ?: return null
    val sha256 = (json.get("sha256") ?: json.get("sha_256"))?.takeIf { ... }?.asString?.trim()...
    val signature = json.get("signature")?.takeIf { ... }?.asString?.trim()...
    if (zipUrl.isBlank() || version.isBlank() || !isSecureRuntimeModuleUrl(zipUrl)) return null
    return RuntimeModuleUpdateInfo(...)
}
```

---

### 2. Checksum Verification (SHA-256)

The `fileSha256Hex()` utility computes a SHA-256 digest of the downloaded module ZIP before installation:

```kotlin
fun fileSha256Hex(file: File): String {
    val digest = MessageDigest.getInstance("SHA-256")
    file.inputStream().use { input ->
        val buffer = ByteArray(8192)
        var read: Int
        while (input.read(buffer).also { read = it } > 0) {
            digest.update(buffer, 0, read)
        }
    }
    return digest.digest().joinToString("") { "%02x".format(it) }
}
```

Supports both `"sha256"` and `"sha_256"` field names in the update JSON for compatibility with different module repository formats.

---

### 3. Install-Time Checksum Check

After downloading via `downloadRuntimeModuleAsset()`, the expected SHA-256 from update JSON is compared against the actual file digest. A mismatch cancels the installation with a checksum error:

```kotlin
val expectedSha256 = target.updateInfo.sha256
if (!expectedSha256.isNullOrBlank()) {
    val actual = withContext(Dispatchers.IO) { DownloadUtils.fileSha256Hex(downloadedFile) }
    if (!actual.equals(expectedSha256.trim(), ignoreCase = true)) {
        installRunning = false
        installSuccess = false
        installLog = installLog + listOf(
            "",
            context.getString(R.string.runtime_module_checksum_mismatch)
        )
        return@launch
    }
}
```

---

### 4. Dedicated Download Helper

`downloadRuntimeModuleAsset()` ensures consistent behavior: downloads go to the configured `Downloads/ABK` directory with `preserveDownloadedZip = true` so the installer can verify the checksum before flashing.

```kotlin
suspend fun downloadRuntimeModuleAsset(
    context: Context,
    token: String?,
    url: String,
    name: String,
    sizeBytes: Long = 0L,
    runId: Long = -2000000001L,
    runTitle: String,
    downloadDirectoryPath: String? = null,
    preserveDownloadedZip: Boolean = true,
): DownloadResult {
    return downloadDirectAsset(
        context = context, token = token, url = url,
        name = name, sizeBytes = sizeBytes,
        runId = runId, runTitle = runTitle,
        sourceAssetId = 0L,
        downloadDirectoryPath = downloadDirectoryPath,
        storageSubdirectory = "ABK",
        preserveDownloadedZip = preserveDownloadedZip,
        bundleWithNotices = false
    )
}
```

---

### 5. Summary of Security Flow

```
Update JSON URL (HTTPS required)
  ↓ fetch & parse
Extract sha256 / sha_256, zipUrl, changelog, version
  ↓ zipUrl must be HTTPS
Download ZIP via downloadRuntimeModuleAsset() → Downloads/ABK/
  ↓
Compute SHA-256 of downloaded file via fileSha256Hex()
  ↓ compare with expected sha256 from update JSON
  ├─ mismatch → reject installation with checksum error
  └─ match    → proceed to installModule() via root shell
```

xingguangcuican6666 (approved):

PR #158, opened 2026-07-19
security(cli): store GitHub credentials securely

### LINKED ISSUE #154: security(cli): avoid storing GitHub tokens in plaintext
## Problem

`abk login` currently writes the GitHub access token directly to the CLI JSON configuration as the `token` field. The config file is restricted to mode `0600` on POSIX, but the credential is still stored as plaintext, and the same permission hardening is not guaranteed on Windows.

A GitHub token may grant access beyond this repository, depending on the scopes selected by the user. Plaintext persistence increases the impact of local malware, unintended backups or synchronization, and accidental configuration disclosure.

## Desired behavior

- Store persistent credentials in the platform's native credential store when available:
  - Windows Credential Manager
  - macOS Keychain
  - Secret Service/libsecret on Linux
- Continue supporting `GITHUB_TOKEN`, `GH_TOKEN`, and `--token` as non-persistent authentication methods.
- Migrate an existing plaintext `token` safely, then remove it from the JSON config after successful migration.
- Make `abk logout` remove credentials from the secure store, the encrypted fallback, and any legacy config entry.
- Never expose the token or derived key material in logs or JSON output.
- Never silently fall back to plaintext storage.

## Machine-bound encrypted fallback

Some systems do not provide a working Secret Service implementation or D-Bus session. Native credential storage must remain preferred, but the CLI should provide a versioned local fallback for these environments:

- Generate 32 cryptographically random bytes on first use and persist them as Base64 in a user-only file.
- Obtain a stable machine identifier without persisting an additional copy of it.
- Derive a 256-bit key with HKDF-SHA-256 from the decoded random value and machine identifier, using an ABK-specific context string.
- Encrypt the credential with AES-256-GCM and a fresh random 96-bit nonce on every write.
- Persist only versioned metadata, the Base64 random value, nonce, ciphertext, and authentication tag; never persist the derived key.
- Re-derive the key and authenticate/decrypt the value on subsequent runs.
- If the machine identifier changes, metadata is corrupt, or authentication fails, fail closed and offer a command or prompt to reset the unusable credential.
- Restrict fallback files to the current user (`0600` on Unix-like systems) and show a clear degraded-storage warning before enabling this fallback.

### Security boundary

The persisted random value and machine identifier are both obtainable by a local process. This fallback protects against direct token disclosure from an accidentally shared config or backup, but it is not equivalent to Secret Service and cannot protect against an attacker already able to read the user's files and machine identifier. Stronger local-attacker protection requires a user passphrase, TPM-backed key, or native credential store.

## Acceptance criteria

- A normal `abk login` does not write the raw GitHub token to the JSON config.
- Native credential storage remains the default whenever it is available.
- Without Secret Service, the fallback survives a process restart and decrypts successfully on the same machine.
- The fallback uses HKDF-SHA-256 and AES-256-GCM with a unique nonce for each write.
- No plaintext credential or raw derived key is persisted or logged.
- Legacy plaintext credentials are migrated safely and erased after successful migration.
- `abk logout` clears credentials from every supported persistence backend.
- Modified ciphertext/tag, corrupt metadata, and machine-identifier mismatch are detected, never return plaintext, and can be reset safely.
- Automated tests cover backend preference, migration, fallback round trips, nonce uniqueness, tamper detection, machine-identifier mismatch, permissions, logout, and reset behavior.
- CLI documentation explains credential storage, degraded fallback limitations, backup/migration implications, and non-persistent authentication.

POSIX mode `0600` should remain as defense in depth for all CLI configuration, but it should not be treated as a substitute for secret storage.

Mazha0309:
## Summary

- Store persistent GitHub tokens in Windows Credential Manager, macOS Keychain, or Linux Secret Service.
- Use AES-256-GCM v2 records backed by a separately stored random local master key when no supported native credential service is available, adding stable-machine binding when possible.
- Use a local-master-only mode when a stable machine identifier is unavailable, with explicit degraded-storage warnings and automatic upgrades to stronger storage.
- Safely migrate legacy plaintext tokens, redact runtime credentials, and clean persistent credential state on logout.
- Use write-ahead recovery markers and verified rollback so interrupted native credential updates fail closed.
- Fail closed when logout cannot authenticate the recorded native provider, preserving retry state and requiring explicit manual cleanup instead of deleting from a guessed provider.
- Let a GitHub-verified interactive login recover from a missing machine identifier or temporarily unavailable native operation while retaining only a revalidated or authenticated deferred-cleanup identity.
- Auto-heal authenticated deferred fallbacks on read when the exact recorded native provider recovers, while still strengthening local-key fallback when only a machine identifier becomes available.
- Keep already-logged-out invocations read-only when no credential state exists, including configurations that contain only a stale lock file.
- Verify legacy plaintext removal only when a parsed legacy token was actually removed, so unrelated malformed configuration cannot reverse a completed credential operation.
- Restore the exact previous secure credential state when legacy plaintext migration cleanup fails, after first restoring and verifying the original configuration.
- Inspect legacy configuration strictly before logout mutates secure storage, rejecting malformed, unreadable, aliased, or non-object files that could otherwise strand plaintext credentials.
- Revalidate plaintext native-provider metadata against the selected system backend before authenticating it into deferred cleanup state, failing closed when that identity cannot be confirmed.
- Initialize the Windows config-lock byte only after acquiring its byte-range lock, eliminating a first-use cross-process race.
- Install the complete Linux Secret Service dependency stack for both source installs and packaged CLI artifacts.

## Scope and compatibility

- Changes are limited to the CLI, CLI tests/documentation, and the ABK CLI packaging workflow.
- Android and Desktop sources and their CI workflows are untouched.
- Command-line and environment tokens remain non-persistent, and existing token precedence is unchanged.
- Existing machine-bound and local-key v1 credential files remain readable; legacy machine-only v1 records migrate to the secret-backed v2 format on normal mutable reads.

## Verification

- 327 CLI tests passed normally.
- Fallback regressions prove that `credentials.json` plus the public machine identifier cannot decrypt v2 records, missing or mismatched master keys fail closed, v2-to-v1 downgrade attempts fail authentication, and authenticated v1 records migrate without losing deferred-provider identity.
- Logout regressions cover provider mismatch, untrusted metadata, clean fallback and missing-primary states that must never guess a native provider, corrupt transaction markers that require a provider-bound primary, interrupted cleanup retries, and native backends that report deletion while retaining the token.
- Credential metadata, transaction-marker, and local-key removals fsync the parent directory; fault-injection tests cover post-unlink sync failure, durable `ENOENT` retries, and truly stateless cleanup.
- Login recovery regressions cover missing versus changed machine identifiers, factory-level versus operation-level native unavailability, provider mismatch before backend access, exact-provider auto-healing, authenticated cleanup-provider metadata, rollback failures, repeated degraded writes, exact-provider logout, and both supported AES-GCM implementations.
- Stateless logout regressions verify that clean and stale-lock-only configurations are not modified, while orphaned credential keys still enter the serialized cleanup path.
- Migration rollback regressions cover removing a newly created secure credential, restoring a previous secure credential, lost cleanup responses, unverifiable configuration rollback, and combined cleanup/rollback failures.
- Strict logout regressions verify that malformed, unreadable, non-object, symlinked, and hardlinked legacy configurations fail before secure deletion, while valid token-free configuration still proceeds normally.
- Cross-process lock regressions verify that the Windows lock byte is initialized only after acquisition and that failed child processes are always reaped before temporary-directory cleanup.
- Malformed-config regressions verify that unrelated invalid JSON is preserved without turning an already verified store or delete into a false failure.
- Source and native Linux install contracts require `SecretStorage>=3.5`, while macOS and Windows remain unchanged.
- All 102 credential-store tests passed with `cryptography` blocked and PyCryptodomex selected; cross-backend v2 and legacy-v1 fixtures also passed.
- Python compilation, i18n catalog parity, and diff checks passed.
- A fresh Linux PyInstaller bundle passed its frozen self-test.
- A Python 3.12 i386 bundle built with pinned `cryptography 49.0.0`, `keyring 25.7.0`, and `SecretStorage 3.5.0`; its frozen self-test and Secret Service archive checks passed. The cached dependency install completed in about three seconds after the verified cold source build.
- The pinned cryptography armv7 abi3 wheel was resolved successfully, avoiding a Rust build under emulation.
- The frozen binary was tested with `/etc/machine-id` masked: legacy migration created authenticated `credentials.json` and `credentials.key` files with `0600` permissions, emitted both degraded-storage warnings, stored no plaintext token, and removed both files on logout.

Closes #154


cli/credential_store.py
```diff
@@ -0,0 +1,1276 @@
...
+            ):
+                raise CredentialCorrupt("native credential identity is invalid")
+            if not include_native:
+                return None
+            backend = self._native_backend_factory()
+            if metadata.get("provider") != backend.name:
+                raise NativeStoreUnavailable(
+                    "the configured native credential provider is unavailable"
+                )
+            token = backend.get()
```
xingguangcuicanbot, line 1055: Location: cli/credential_store.py:773

Issue: The new native-storage path validates `backend.get()` during `store()`, but `read()` returns whatever the backend supplies without checking its type. If a keyring backend regresses or returns bytes/another object after an external change, `get_token()` will pass that invalid value upward instead of failing closed, which breaks the credential-integrity guarantees this PR is adding and can lead to malformed authentication requests.

Suggestion: Validate native read results the same way `store()` validates `previous_token`: accept `None` or `str`, and raise `NativeStoreError` for anything else.

```suggestion
token = backend.get()
if token is not None and not isinstance(token, str):
    raise NativeStoreError(
        f"{backend.name} returned an invalid GitHub credential"
    )
self._remove_unused_local_key()
return token
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
One correctness issue found in the native credential read path.

### Decision
request_changes

### Notes

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The credential-store implementation is largely well tested, but the packaging workflow still regresses native credential support in shipped CLI artifacts.

### Findings
- [medium] .github/workflows/build-abk-cli.yml:312 (https://github.com/xingguangcuican6666/ABK/blob/3f8b9fe3cf2a0fae5d5d1166855d00f37529ba27/.github/workflows/build-abk-cli.yml#L312)
  - Issue: The Linux packaging jobs for `linux-armhf` and `linux-x86` install `pyinstaller certifi pycryptodome pynacl` but never install `keyring`. Those binaries therefore cannot import `keyring.backends.SecretService` at runtime, so `create_native_backend()` will always fall back to the degraded file-based store even on systems with a working Secret Service session. That contradicts the PR's stated behavior for supported Linux platforms and creates an architecture-specific regression in released artifacts.
  - Suggestion: Install `keyring` in the cross-packaging containers as well, so the bundled Linux x86/armhf binaries can actually use Secret Service on supported hosts.
- [medium] .github/workflows/build-abk-cli.yml:216 (https://github.com/xingguangcuican6666/ABK/blob/3f8b9fe3cf2a0fae5d5d1166855d00f37529ba27/.github/workflows/build-abk-cli.yml#L216)
  - Issue: The PyInstaller commands only add hidden imports for cryptography/PyCryptodome, but `credential_store.create_native_backend()` imports `keyring.backends.Windows`, `keyring.backends.macOS`, and `keyring.backends.SecretService` dynamically at runtime. PyInstaller does not reliably discover those backend modules from dynamic imports, so the frozen CLI can pass `self-test` yet still fail to load any native credential backend and silently downgrade to fallback storage. This affects the packaged binaries even when `keyring` is installed during the build.
  - Suggestion: Collect the keyring backend modules in the frozen bundle, for example with `--collect-submodules keyring` or explicit hidden imports for the supported backends.

PR URL: https://github.com/xingguangcuican6666/ABK/pull/158

.github/workflows/build-abk-cli.yml
```diff
@@ -149,7 +149,15 @@ jobs:
...
           python -m pip install --upgrade pip
           python -m pip install pyinstaller certifi pycryptodome pynacl
-          python -m pip install cryptography 2>/dev/null || echo "::warning::cryptography unavailable, using pycryptodome"
+          cryptography_available=false
+          if python -m pip install cryptography 2>/dev/null; then
+            cryptography_available=true
+          else
+            echo "::warning::cryptography unavailable, using pycryptodome"
+          fi
+          if [[ "$RUNNER_OS" == "macOS" || "$cryptography_available" == "true" ]]; then
```
xingguangcuicanbot: Location: .github/workflows/build-abk-cli.yml:158

Issue: On non-Windows builds, `keyring` is only installed when `cryptography` was installed successfully. Those two concerns are independent: the CLI can use `pycryptodome` for AES-GCM and still use Secret Service for native credential storage. As written, any Linux build that falls back to `pycryptodome` will silently lose native Secret Service support and always drop to the weaker file-based credential backend, even on machines where a supported native store is present.

Suggestion: Install `keyring` unconditionally for non-Windows builds, and keep the `cryptography` fallback logic only for the AES backend selection.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The secure-storage implementation looks solid overall, but the packaging workflow currently ships some Linux artifacts without the native credential-store dependency, which forces a weaker fallback backend even on systems that support Secret Service.

### Decision
request_changes

### Notes

- .github/workflows/build-abk-cli.yml:312 The cross-packaging container for `linux-armhf` omits `keyring` entirely (the same omission is repeated in the `linux-x86` block). That means the released cross-built Linux binaries cannot use Secret Service at all, so they will always persist tokens via the degraded file-encryption fallback. This is a behavior regression for those official artifacts, not just reduced test coverage.

cli/credential_store.py
```diff
@@ -0,0 +1,1280 @@
...
+        ):
+            return self._delete_unusable_metadata()
+        try:
+            backend = self._native_backend_factory()
+        except NativeStoreUnavailable:
+            # A valid native marker identifies a credential that still needs
+            # cleanup. Keep it so a later logout can retry the fixed account.
+            raise
+        if metadata.get("provider") != backend.name:
+            return self._delete_unusable_metadata(backend=backend)
```
xingguangcuicanbot: Location: cli/credential_store.py:1208

Issue: When the on-disk metadata says the native provider is different from the currently available provider, `delete()` still falls through to `_delete_unusable_metadata(backend=backend)`. That helper immediately deletes the current provider's `ABK CLI/github.com` entry, even though the trusted metadata explicitly says the stored credential belongs to some other provider. In a migration or copied-config scenario, `abk logout` can therefore remove the wrong token and then delete the metadata that would have allowed the original provider to be cleaned up later.

Suggestion: Fail closed on provider mismatch and keep the metadata for a later retry against the recorded provider.

```suggestion
        if metadata.get("provider") != backend.name:
            raise NativeStoreUnavailable(
                "the configured native credential provider is unavailable"
            )
```

cli/credential_store.py
```diff
@@ -0,0 +1,1280 @@
...
+            or pending_removed
+        )
+
+    def _delete_unusable_metadata(self, *, backend=None):
+        """Reset unreadable metadata while best-effort cleaning fixed native state."""
+        native_removed = False
+        cleanup_error = None
+        try:
+            backend = backend or self._native_backend_factory()
+            native_removed = backend.delete()
```
xingguangcuicanbot: Location: cli/credential_store.py:1265

Issue: `_delete_unusable_metadata()` deletes the current native backend entry even when the metadata is unreadable or unsupported. At that point the file no longer provides a trustworthy binding to a specific provider or state, so cleanup against the current backend is guesswork. A corrupt `credentials.json` copied from another machine can therefore cause `abk logout` to delete this machine's valid native credential unexpectedly.

Suggestion: Do not touch the native backend when the metadata cannot be trusted to identify it. Limit recovery here to removing local metadata/key files, and surface an error if manual native cleanup may still be required.

```suggestion
        # Untrusted metadata cannot safely identify a native credential.
        native_removed = False
        cleanup_error = None
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The new credential store logic is mostly thorough, but there are two fail-closed violations in logout cleanup that can delete the wrong native credential and orphan the recorded one.

### Decision
request_changes

### Notes

cli/credential_store.py
```diff
@@ -0,0 +1,1280 @@
...
+            raise CredentialStoreError("the GitHub credential is empty")
+        if len(token.encode("utf-8")) > MAX_TOKEN_SIZE:
+            raise CredentialStoreError("the GitHub credential is too large")
+        if self._read_pending_metadata() is not None:
+            raise NativeRollbackError("native credential cleanup is pending")
+        existing_metadata = self._read_metadata()
+        if existing_metadata is not None:
+            existing_backend = existing_metadata.get("backend")
+            if existing_backend in FALLBACK_BACKENDS:
+                self._decrypt_fallback(existing_metadata)
```
xingguangcuicanbot: Location: cli/credential_store.py:1051

Issue: When the existing persisted credential uses the machine-bound fallback, `store()` always decrypts that old record before deciding how to write the new token. If the machine identifier has become unavailable (for example, `/etc/machine-id` is masked, which this PR explicitly supports for fresh logins), `_decrypt_fallback()` raises and the new token is never stored. In practice, `abk login` cannot recover from a previously machine-bound credential in the same degraded environment; the user must manually `logout` first.

Suggestion: Allow an explicit overwrite path for an existing machine-bound record when the caller already has a fresh token but the old record is unreadable only because the machine identifier is unavailable. The store logic should be able to reselect a backend and replace the old metadata instead of hard-failing during preflight validation.

cli/credential_store.py
```diff
@@ -0,0 +1,1280 @@
...
+                    )
+            elif existing_backend == "native":
+                self.read(include_native=False)
+            else:
+                raise CredentialCorrupt("credential backend is unsupported")
+        try:
+            backend = self._native_backend_factory()
+            previous_token = backend.get()
+        except NativeStoreUnavailable:
+            if (
```
xingguangcuicanbot: Location: cli/credential_store.py:1064

Issue: If the current persisted credential is native-backed and the OS keyring is unavailable, this branch re-raises `NativeStoreUnavailable` instead of falling back to encrypted-file storage for the new token. That leaves users with no recovery path: `login` cannot persist the fresh token, and `logout` also fails while the native backend is down. The PR description promises fallback when no supported native credential service is available, but existing native metadata currently blocks that fallback.

Suggestion: Handle native-backend unavailability during a new store as a recoverable downgrade case. Persist the fresh token into the fallback backend while preserving enough state to retry native cleanup later, rather than requiring the old keyring service to come back before any new login can succeed.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The credential-store implementation is close, but there are two recovery gaps where a user cannot replace a broken persisted token with a fresh login token. Those cases can strand users until they manually delete storage artifacts or restore the old backend.

### Decision
request_changes

### Notes

cli/requirements.txt
```diff
@@ -1,3 +1,4 @@
 cryptography>=41
 PyNaCl>=1.5
 certifi>=2024.2.2
+keyring>=25.7
```
xingguangcuicanbot, line 4: Location: cli/requirements.txt:4

Issue: On Linux, `create_native_backend()` now hard-codes `keyring.backends.SecretService`, but `requirements.txt` only adds `keyring`. The Secret Service backend still needs the `SecretStorage` stack, so a user who follows the documented `pip install -r cli/requirements.txt` flow will not actually get the native keyring backend and will always fall back to degraded local file storage. That is a correctness regression relative to the PR's stated behavior, and it weakens the storage guarantees for the default Linux installation path.

Suggestion: Add the Linux-only Secret Service dependency set to the runtime install path, and mirror that in the native Linux packaging step/README so the default install really enables the advertised backend.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The credential-store implementation is largely well covered, but the Linux dependency wiring is incomplete and causes the new Secret Service path to silently degrade to file-based fallback for normal source installs.

### Decision
request_changes

### Notes

cli/credential_store.py
```diff
@@ -0,0 +1,1466 @@
...
+                native_removed = False
+            else:
+                if (
+                    cleanup_provider is not None
+                    and backend.name != cleanup_provider
+                ):
+                    raise NativeStoreUnavailable(
+                        "the deferred native credential provider is unavailable"
+                    )
+                native_removed = self._delete_native_and_verify(backend)
```
xingguangcuicanbot, line 1612: Location: cli/credential_store.py:1368

Issue: When the active record is a fallback file and `native_cleanup_pending` is `False`, `delete()` still deletes whatever token is currently present in the native backend. That backend was never recorded in the authenticated fallback metadata, so this can wipe an unrelated native token that appeared later, for example after keyring became available again or after another install populated the OS store. This conflicts with the PR's fail-closed and exact-provider design and makes `abk logout` destructive against guessed providers.

Suggestion: Only delete from the native backend when the fallback metadata explicitly carries an authenticated `native_cleanup_provider`. If there is no deferred provider in the record, remove only the fallback files.

```suggestion
else:
    if cleanup_provider is None:
        native_removed = False
    elif backend.name != cleanup_provider:
        raise NativeStoreUnavailable(
            "the deferred native credential provider is unavailable"
        )
    else:
        native_removed = self._delete_native_and_verify(backend)
```

cli/credential_store.py
```diff
@@ -0,0 +1,1466 @@
...
+            or not metadata["provider"]
+            or metadata.get("service") != CREDENTIAL_SERVICE
+            or metadata.get("account") != CREDENTIAL_ACCOUNT
+        ):
+            raise CredentialCorrupt(
+                "native credential transaction metadata cannot identify its "
+                "provider safely"
+            ) from marker_error
+        try:
+            self._ensure_pending_primary_compatible(metadata)
```
xingguangcuicanbot: Location: cli/credential_store.py:1411

Issue: `_delete_corrupt_pending()` will delete the current native credential as long as the damaged pending file contains a plausible provider, service, and account triplet, even when there is no valid primary metadata tying that marker to an ABK-managed credential. A stray or malformed `credentials.pending.json` can therefore cause `abk logout` to erase the live native token without any authenticated local state proving ownership.

Suggestion: Require a valid primary metadata record that matches the same provider before using a corrupt pending marker to clean native state. If that proof is missing, keep the marker and fail closed with a manual-cleanup error instead of deleting the native credential.

```suggestion
try:
    primary_metadata = self._read_metadata()
    if primary_metadata is None:
        raise CredentialCorrupt(
            "native credential transaction metadata cannot be verified without matching primary metadata"
        )
    self._ensure_pending_primary_compatible(metadata)
    backend = self._native_backend_factory()
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
I found two correctness/security regressions in the new credential cleanup paths. Both let `logout` delete a native GitHub credential without an authenticated provider binding from the current on-disk state.

### Decision
request_changes

### Notes

cli/credential_store.py
```diff
@@ -0,0 +1,1478 @@
...
+                "a native cleanup provider requires pending cleanup"
+            )
+        encoded_token = token.encode("utf-8")
+        if not encoded_token or len(encoded_token) > MAX_TOKEN_SIZE:
+            raise CredentialStoreError("the GitHub credential has an invalid size")
+        seed = secrets.token_bytes(_SEED_SIZE)
+        nonce = secrets.token_bytes(_NONCE_SIZE)
+        if backend == FALLBACK_BACKEND:
+            machine_id = machine_id or self._validated_machine_id()
+            key = _hkdf_sha256(seed, machine_id)
```
xingguangcuicanbot: Location: cli/credential_store.py:714

Issue: The fallback encryption key is derived only from the per-record `seed` stored in `credentials.json` and the machine identifier. That machine identifier is not a secret on the supported platforms, so anyone who can copy the credential file and read the host's machine ID can decrypt the GitHub token offline. In practice this means the advertised 'secure' fallback is only obfuscation, not confidentiality, and it materially weakens the security guarantees of the feature.

Suggestion: Do not use a key derivation path that depends only on public or file-local inputs. Either fail closed when no native credential store is available, or derive the fallback key from a secret that is not stored alongside the ciphertext (for example, a separately protected local master key, or an OS-protected secret). At minimum, the machine-bound mode should not be presented as secure token storage.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The new native-backend integration is thorough, but the machine-bound fallback does not actually keep the token secret, so the PR should not be merged as-is.

### Decision
request_changes

### Notes

cli/credential_store.py
```diff
@@ -0,0 +1,1659 @@
...
+        )
+
+    def _write_pending_metadata(self, metadata):
+        self._write_json_file(
+            self.pending_path,
+            metadata,
+            prefix=".credential-pending-",
+        )
+
+    def _remove_metadata(self):
```
xingguangcuicanbot: Location: cli/credential_store.py:643

Issue: The new WAL-style credential flow fsyncs directory entries after writes, but the deletion helpers do not fsync the parent directory after unlinking `credentials.json`, `credentials.pending.json`, or `credentials.key`. On POSIX filesystems, a power loss or crash after `unlink()` can lose the directory-entry removal even though the method already returned success. In practice that can resurrect a stale pending marker after a successful native-store commit, or bring back credential files after `abk logout`, leaving the CLI stuck in `cleanup pending` or restoring secrets that were reported as deleted. Because this PR explicitly relies on crash-safe rollback and fail-closed cleanup, the delete path needs the same durability guarantee as the write path.

Suggestion: After each successful unlink, fsync the containing directory before returning. Apply the same fix to `_remove_metadata`, `_remove_pending_metadata`, and `_remove_local_key` so transaction-marker removal and logout cleanup are durable across crashes.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The PR substantially improves credential storage, but the new transaction/cleanup implementation still has a crash-consistency hole during file deletion.

### Decision
request_changes

### Notes

cli/credential_store.py
```diff
@@ -0,0 +1,1676 @@
...
+    def read(self, *, include_native=True):
+        if self._read_pending_metadata() is not None:
+            raise NativeRollbackError("native credential cleanup is pending")
+        metadata = self._read_metadata()
+        if metadata is None:
+            return None
+        backend_name = metadata.get("backend")
+        if backend_name in FALLBACK_BACKENDS:
+            token = self._decrypt_fallback(metadata)
+            if metadata["native_cleanup_pending"]:
```
xingguangcuicanbot: Location: cli/credential_store.py:1008

Issue: Fallback records with `native_cleanup_pending=True` return immediately after decryption and never run the upgrade paths used for normal fallback records. That means a token written while the native provider or machine identifier was temporarily unavailable will stay permanently degraded until the user logs in again, and any stale native credential recorded in `native_cleanup_provider` is never auto-healed/cleaned up on later reads even after the original provider recovers. This conflicts with the documented automatic upgrade behavior and leaves recovered systems stuck in the weaker state indefinitely.

Suggestion: In the `native_cleanup_pending` branch, attempt the same strengthening steps used by the normal fallback path: first try to promote back to native storage when the current backend name matches `native_cleanup_provider`, and otherwise opportunistically rebind local-key fallback to the machine-bound format when a stable machine identifier becomes available.

cli/abk.py
```diff
@@ -472,14 +473,208 @@ def save_config(config):
...
+    if not migrate_legacy and not include_native:
+        # Secret collection is best-effort and must not create or wait on the
+        # config lock merely to render an error. Metadata writes are atomic,
+        # and this path cannot migrate or upgrade a backend.
+        return _credential_store().read(include_native=False)
+    with _config_process_lock():
+        return _credential_store().read(include_native=include_native)
+
+
+def _delete_persisted_token():
```
xingguangcuicanbot, line 756: Location: cli/abk.py:635

Issue: `_delete_persisted_token()` now always acquires `_config_process_lock()`, which creates `CONFIG_DIR` and `.config.lock` even when the user is already logged out and no credential files exist. This regresses `abk logout` from a no-op into a write operation: it leaves new state behind on clean machines, and it will now fail in read-only or permission-restricted home directories where the previous implementation succeeded.

Suggestion: Short-circuit `logout` before taking the process lock when there is no config file and no credential metadata/key/pending files to clean up.

```suggestion
credential_paths = (
    CONFIG_FILE,
    CONFIG_DIR / credential_store.CREDENTIAL_FILE_NAME,
    CONFIG_DIR / credential_store.CREDENTIAL_PENDING_FILE_NAME,
    CONFIG_DIR / credential_store.CREDENTIAL_KEY_FILE_NAME,
)
if not any(path.exists() for path in credential_paths):
    return False, None

with _config_process_lock():
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The new credential store is well-tested, but I found two regressions that should be addressed before merging.

### Decision
request_changes

### Notes

cli/abk.py
```diff
@@ -472,14 +473,228 @@ def save_config(config):
...
+            )
+        config = load_config()
+        if "token" in config:
+            config.pop("token", None)
+            try:
+                save_config(config)
+            except OSError as exc:
+                raise credential_store.CredentialStoreError(
+                    "legacy plaintext credential could not be removed"
+                ) from exc
```
xingguangcuicanbot: Location: cli/abk.py:576

Issue: `_verify_legacy_credential_removed()` is called unconditionally after the new credential has already been stored and verified. Because `load_config()` intentionally swallows JSON parse errors, an unrelated malformed `config.json` now makes `login` fail *after* the token was successfully persisted, and the same pattern appears again in `_delete_persisted_token()`. That turns a successful credential-store operation into a false negative for callers and can leave automation retrying an already-completed login/logout.

Suggestion: Only run the post-removal verification when a legacy plaintext token was actually removed, and apply the same guard in `_delete_persisted_token()`.

```suggestion
config = load_config()
if "token" in config:
    config.pop("token", None)
    try:
        save_config(config)
    except OSError as exc:
        raise credential_store.CredentialStoreError(
            "legacy plaintext credential could not be removed"
        ) from exc
    _verify_legacy_credential_removed()
return result
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The secure credential-store implementation is substantial and generally well covered, but there is at least one regression: malformed unrelated `config.json` content can now cause `login`/`logout` to report failure even after the credential operation already succeeded.

### Decision
request_changes

### Notes

cli/abk.py
```diff
@@ -472,14 +473,226 @@ def save_config(config):
...
+            not isinstance(persisted, str)
+            or not hmac.compare_digest(persisted, token)
+        ):
+            raise credential_store.CredentialStoreError(
+                "the stored GitHub credential failed disk verification"
+            )
+        config = load_config()
+        if "token" in config:
+            config.pop("token", None)
+            try:
```
xingguangcuicanbot: Location: cli/abk.py:571

Issue: `_store_persisted_token()` writes and verifies the new secure credential before removing the legacy plaintext token. If `save_config()` or `_verify_legacy_credential_removed()` fails, the function returns an error but leaves the newly stored credential in place. That means a failed login/migration can still persist a valid token and future commands will treat the user as logged in, which is not fail-closed and can leave duplicate credentials behind.

Suggestion: Rollback the just-written secure credential when legacy plaintext cleanup fails, and surface a combined error if the rollback itself cannot be verified.

cli/abk.py
```diff
@@ -472,14 +473,226 @@ def save_config(config):
...
+
+    with _config_process_lock():
+        removed = False
+        error = None
+        try:
+            removed = _credential_store().delete()
+        except credential_store.CredentialStoreError as exc:
+            error = exc
+
+        if CONFIG_FILE.exists():
```
xingguangcuicanbot: Location: cli/abk.py:664

Issue: This logout cleanup path relies on `load_config()`, which silently converts unreadable or malformed `config.json` into `{}`. If a damaged legacy config still contains a plaintext `token`, `abk logout` will report success here without removing or flagging that secret, leaving the token on disk while `loggedIn` becomes `false`. Use a strict read/parse step for legacy-token cleanup so malformed config is treated as a credential-storage error instead of a successful logout.

Suggestion: Do not use `load_config()` for legacy token removal. Parse `config.json` strictly in the cleanup path and fail the operation if the file cannot be authenticated as token-free.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The secure-token migration/logout flow still has two fail-closed gaps around legacy plaintext cleanup: failed migration can leave a new secure token stored despite reporting failure, and malformed legacy config can make logout report success while a plaintext token remains on disk.

### Decision
request_changes

### Notes

cli/credential_store.py
```diff
@@ -0,0 +1,1682 @@
...
+                    if not allow_recovery:
+                        raise NativeRollbackError(
+                            "native credential cleanup is pending"
+                        )
+                    required_native_provider = existing_metadata[
+                        "native_cleanup_provider"
+                    ]
+            elif existing_backend == "native":
+                self.read(include_native=False)
+                required_native_provider = existing_metadata["provider"]
```
xingguangcuicanbot: Location: cli/credential_store.py:1412

Issue: `store(..., allow_recovery=True)` copies `existing_metadata["provider"]` from plaintext native metadata into `required_native_provider` after only calling `read(include_native=False)`, which validates shape but does not authenticate that provider. If `credentials.json` is tampered while the real native backend is unavailable, the next verified login will persist a fallback record whose `native_cleanup_provider` is attacker-controlled. That turns untrusted local metadata into an authenticated cleanup binding, so later auto-heal/logout can target the wrong backend or permanently lose the ability to clean up the real stale native credential.

Suggestion: Do not carry a native cleanup provider forward from native metadata unless the provider has been revalidated against the actual backend. If the backend is unavailable, fail closed instead of persisting a deferred provider derived only from local JSON.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The recovery flow currently authenticates an untrusted provider name from plaintext native metadata, which can bind future cleanup to the wrong credential backend.

### Decision
request_changes

### Notes

cli/credential_store.py
```diff
@@ -0,0 +1,1716 @@
...
+    def _translate(self, operation, exc):
+        unavailable = tuple(
+            error
+            for error in (
+                getattr(self._errors, "NoKeyringError", None),
+                getattr(self._errors, "InitError", None),
+            )
+            if isinstance(error, type)
+        )
+        if unavailable and isinstance(exc, unavailable):
```
xingguangcuicanbot, line 130: Location: cli/credential_store.py:130

Issue: `KeyringBackend._translate()` only treats `keyring.errors.NoKeyringError` and `InitError` as availability failures. In practice, the Secret Service backend also raises plain `OSError` subclasses for transport/session problems; for example, `Keyring().get_password(...)` on a host without a usable D-Bus session raises `PermissionError: [Errno 1] Operation not permitted`. Those exceptions currently become `NativeStoreError`, so `store(..., allow_recovery=True)` never takes the fallback path and interactive login/logout regress from “degraded but usable” to hard failure whenever the native service is temporarily unreachable.

Suggestion: Classify OS-level backend transport failures as `NativeStoreUnavailable` as well, so temporary keychain/session outages still trigger the documented fallback and recovery behavior.

```suggestion
        unavailable = tuple(
            error
            for error in (
                getattr(self._errors, "NoKeyringError", None),
                getattr(self._errors, "InitError", None),
                OSError,
            )
            if isinstance(error, type)
        )
        if unavailable and isinstance(exc, unavailable):
            raise NativeStoreUnavailable(
                f"{self.name} is unavailable"
            ) from exc
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The secure-storage rewrite is generally well covered, but the keyring adapter currently misclassifies real native-backend availability failures as hard store errors. That blocks the intended fallback/recovery flow on Linux Secret Service outages and should be fixed before merge.

### Decision
request_changes

### Notes

cli/credential_store.py
```diff
@@ -0,0 +1,1716 @@
...
+                raise CredentialStoreError(
+                    "native cleanup safety marker failed verification"
+                )
+        except (CredentialStoreError, OSError) as exc:
+            raise CredentialCorrupt(
+                "untrusted credential metadata could not be safely reset"
+            ) from exc
+        return self._complete_cleanup_blocked_reset(metadata_error)
+
+    def _complete_cleanup_blocked_reset(self, metadata_error=None):
```
xingguangcuicanbot, line 1704: Location: cli/credential_store.py:1704

Issue: `_complete_cleanup_blocked_reset()` never removes `credentials.pending.json`, so once `_delete_unusable_metadata()` writes the cleanup-blocked marker, every later `delete()` call will hit the same branch and fail forever even after the native secret is manually cleaned up. That leaves the store in a permanent unrecoverable state.

Suggestion: Add a recovery path that clears the cleanup-blocked marker after manual native cleanup has been verified; otherwise the marker must not be written as a permanent state.

cli/abk.py
```diff
@@ -472,14 +474,349 @@ def save_config(config):
...
+        # creating the configuration directory solely for its lock file.
+        return False, None
+
+    with _config_process_lock():
+        try:
+            config = _load_config_for_credential_cleanup()
+        except credential_store.CredentialStoreError as exc:
+            return False, exc
+        removed = False
+        error = None
```
xingguangcuicanbot, line 783: Location: cli/abk.py:785

Issue: This logout path deletes the secure credential first and only then tries to remove the legacy plaintext config. If `save_config()` or the verification step fails, the code records an error but never restores the deleted secure token, so logout can fail halfway through and lose the previous secure state.

Suggestion: Rollback the secure credential if legacy config cleanup fails, so logout remains atomic instead of leaving a half-removed credential behind.

```suggestion
        store = _credential_store()
        previous_token = None
        try:
            previous_token = store.read()
            removed = store.delete()
        except credential_store.CredentialStoreError as exc:
            error = exc

        if "token" in config:
            try:
                config.pop("token", None)
                save_config(config)
                _verify_legacy_credential_removed()
                removed = True
            except (OSError, credential_store.CredentialStoreError) as exc:
                cleanup_error = credential_store.CredentialStoreError(
                    "legacy plaintext credential could not be removed"
                )
                error = cleanup_error if error is None else credential_store.CredentialStoreError(
                    f"{error}; {cleanup_error}"
                )
                if previous_token is not None:
                    try:
                        _rollback_secure_credential(store, previous_token, args=args)
                    except credential_store.CredentialStoreError as rollback_error:
                        error = credential_store.CredentialStoreError(
                            f"{error}; secure credential rollback failed: {rollback_error}"
                        )
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
Two regressions: the cleanup-blocked marker is permanently unrecoverable, and logout can delete the secure credential before legacy cleanup succeeds.

### Decision
request_changes

### Notes

PR #165, opened 2026-07-23
fix(app): Move kernel options into build child page

### LINKED ISSUE #161: “自定义内核选项”的条目过多时的折叠和一键清除
### Feature Description / 功能描述

“自定义内核选项”的条目过多时的折叠和一键清除

### Use Case / 使用场景

刚刚发现导入了一堆选项之后太长了，没折叠想点开始编译有些麻烦，或者想全清掉的时候一个一个点太麻烦

### Additional Info / 补充信息

_No response_

xingguangcuican6666:


app/src/main/java/com/abk/kernel/ui/screens/BuildScreen.kt
```diff
@@ -1839,6 +1916,22 @@ fun BuildScreen(
                                 .padding(padding)
                                 .fillMaxSize()
                         )
+                    } else if (showKernelOptionsPage) {
+                        BuildKernelOptionsPage(
+                            padding = padding,
```
xingguangcuicanbot, line 1931: Location: app/src/main/java/com/abk/kernel/ui/screens/BuildScreen.kt:1932

Issue: The new child page reuses the main build screen's `scrollBehavior` for the kernel-options `LazyColumn`, but the overlay `ExpressiveTopBar` does not use that same behavior. As a result, scrolling inside the kernel-options page mutates the underlying build screen app-bar state, so when the user closes the overlay the main page can come back unexpectedly collapsed/offset. This is a regression introduced by moving the list into a child page. Use a dedicated scroll behavior for the child page (wired into both the child top bar and the list), or remove the nested-scroll hookup there.

Suggestion: Do not pass the root screen's `scrollBehavior` into `BuildKernelOptionsPage`. Create a separate child-page `TopAppBarScrollBehavior` and use it consistently inside the overlay, or remove nested scrolling from the child page.

app/src/main/java/com/abk/kernel/ui/screens/BuildScreen.kt
```diff
@@ -242,9 +281,23 @@ fun BuildScreen(
     fun openBuildQueuePage() {
```
xingguangcuicanbot, line 282: Location: app/src/main/java/com/abk/kernel/ui/screens/BuildScreen.kt:288

Issue: `kernelOptionSearchQuery` is kept in saveable state and `openKernelOptionsPage()` does not reset it. After the user closes the page, imports/applies a different plan, or bulk-deletes options, reopening the child page can still show an old filter and therefore an apparently empty/incomplete list even though the summary card says options exist. On the old inline UI, the full list was always shown by default, so this is a user-visible regression that can look like data loss.

Suggestion: Clear the saved filter when entering or leaving the kernel-options child page so the page always opens on the full option set.

```suggestion
fun openKernelOptionsPage() {
    childPageBack.resetProgress()
    showPlanLibraryPage = false
    showBuildQueuePage = false
    kernelOptionSearchQuery = ""
    showKernelOptionsPage = true
}
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The PR is generally coherent, but the kernel-options child page introduces two regressions: it leaks scroll state into the parent build screen, and it preserves stale search filters across reopenings in a way that can hide existing options.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/screens/BuildScreen.kt
```diff
@@ -203,7 +215,30 @@ fun BuildScreen(
     val customModuleGroups = remember(config.customExternalModules, catalogModuleByUrl) {
         groupBuildCustomExternalModules(config.customExternalModules, catalogModuleByUrl)
     }
-    val childPageVisible = showPlanLibraryPage || showBuildQueuePage
+    val kernelOptionSummary = summarizeCustomKernelOptions(config.customKernelOptions)
+    val kernelOptionModeEnabledYLabel = stringResource(R.string.build_kernel_option_mode_y)
+    val kernelOptionModeEnabledMLabel = stringResource(R.string.build_kernel_option_mode_m)
+    val kernelOptionModeDisabledLabel = stringResource(R.string.build_kernel_option_mode_disabled)
+    val kernelOptionModeIgnoreLabel = stringResource(R.string.build_kernel_option_mode_ignore)
```
xingguangcuicanbot, line 222: Location: app/src/main/java/com/abk/kernel/ui/screens/BuildScreen.kt:227

Issue: The filter normalizes each option's searchable text with `Locale.ROOT`, but the user's query is lowercased with the device default locale. On locales with special case mappings such as Turkish, searching for symbols containing `I` (for example `CONFIG_IPV6`) stops matching even though the entry exists. Use the same locale for both sides of the comparison.

Suggestion: Normalize the search query with `Locale.ROOT` so it matches the locale-independent normalization already used in `buildCustomKernelOptionSearchText`.

```suggestion
val query = kernelOptionSearchQuery.trim().lowercase(Locale.ROOT)
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The PR is close, but the new kernel-option search introduces a locale-sensitive matching regression that will break searches on some devices.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/screens/BuildScreen.kt
```diff
@@ -2583,6 +2678,221 @@ private fun formatCustomKernelImportSummary(context: Context, result: CustomKern
...
+            BuildKernelOptionSearchField(
+                value = searchQuery,
+                onValueChange = onSearchQueryChange
+            )
+        }
+        item(key = "summary") {
+            Text(
+                text = buildCustomKernelOptionSummaryText(summary),
+                style = MaterialTheme.typography.bodySmall,
+                color = MaterialTheme.colorScheme.onSurfaceVariant
```
xingguangcuicanbot, line 2717: Location: app/src/main/java/com/abk/kernel/ui/screens/BuildScreen.kt:2722

Issue: `LazyColumn.items` now uses `indexed.value.symbol` as the item key. Custom kernel options are not globally deduplicated today: importing options appends onto the existing list, so the same symbol can appear more than once. When that happens, Compose receives duplicate keys and can throw or mis-associate row state as soon as the kernel-options page opens.

Suggestion: Use a key that stays unique even when two entries share the same symbol, for example the original list index combined with the symbol.

```suggestion
items(
    items = options,
    key = { indexed -> "${indexed.index}:${indexed.value.symbol}" }
) { indexed ->
```

app/src/main/java/com/abk/kernel/viewmodel/MainViewModel.kt
```diff
@@ -5764,6 +5785,41 @@ internal fun parseCustomKernelOptionsText(text: String): CustomKernelOptionsImpo
...
+}
+
+internal fun removeCustomKernelOptions(
+    options: List<CustomKernelOption>,
+    symbols: Collection<String>
+): List<CustomKernelOption> {
+    if (options.isEmpty() || symbols.isEmpty()) return options
+    val normalizedTargets = symbols.mapNotNull { raw ->
+        KernelSupport.normalizeCustomKernelSymbol(raw).takeIf { it.isNotBlank() }?.lowercase()
+    }.toSet()
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/viewmodel/MainViewModel.kt:5818

Issue: The new bulk-delete path removes entries by symbol, not by the filtered rows the user actually selected. If the config contains duplicate symbols with different modes/raw values, clearing a filtered search will delete every occurrence of that symbol, including hidden rows that did not match the current filter. This is a regression introduced by the new filtered-clear workflow.

Suggestion: Preserve row identity for bulk deletion, for example by deleting by original indices or by passing a stable per-entry ID instead of only the normalized symbol.

app/src/main/java/com/abk/kernel/viewmodel/MainViewModel.kt
```diff
@@ -5764,6 +5785,41 @@ internal fun parseCustomKernelOptionsText(text: String): CustomKernelOptionsImpo
...
+        disabled = disabled,
+        ignored = ignored
+    )
+}
+
+internal fun removeCustomKernelOptions(
+    options: List<CustomKernelOption>,
+    symbols: Collection<String>
+): List<CustomKernelOption> {
+    if (options.isEmpty() || symbols.isEmpty()) return options
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/viewmodel/MainViewModel.kt:5815

Issue: `lowercase()` here uses the device default locale. On locales with special casing rules such as Turkish, symbol matching for bulk delete can fail because `CONFIG_*` identifiers are ASCII protocol data, not natural-language text. The new helper should normalize with `Locale.ROOT` instead.

Suggestion: Use locale-independent casing when normalizing kernel-option symbols for comparison.

```suggestion
val normalizedTargets = symbols.mapNotNull { raw ->
    KernelSupport.normalizeCustomKernelSymbol(raw)
        .takeIf { it.isNotBlank() }
        ?.lowercase(java.util.Locale.ROOT)
}.toSet()
if (normalizedTargets.isEmpty()) return options
return options.filterNot { option ->
    KernelSupport.normalizeCustomKernelSymbol(option.symbol)
        .lowercase(java.util.Locale.ROOT) in normalizedTargets
}
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The PR introduces a useful kernel-options child page, but the new list and bulk-delete flows have correctness issues: duplicate symbols can break the `LazyColumn` key contract, filtered bulk delete can remove hidden duplicates, and the new symbol matcher uses locale-sensitive casing.

### Decision
request_changes

### Notes

xingguangcuicanbot (approved):
## Automated PR Review

### Summary
No actionable issues found.

### Decision
approve

### Notes
Codex recommendation: approve

PR #172, opened 2026-07-25
fix(app): restore fork signing key downloads

xingguangcuican6666:


app/src/test/java/com/abk/kernel/data/repository/GitHubRepositoryNetworkTest.kt
```diff
@@ -0,0 +1,120 @@
...
+                      "fork": false,
+                      "default_branch": "main",
+                      "private": false,
+                      "html_url": "https://github.com/owner/repo"
+                    }
+                    """.trimIndent()
+                )
+        )
+
+        val result = repository.getRepo("owner", "repo")
```
xingguangcuicanbot, line 56: Location: app/src/test/java/com/abk/kernel/data/repository/GitHubRepositoryNetworkTest.kt:56

Issue: This test calls `repository.getRepo(...)`, but `GitHubRepository` does not expose any `getRepo` method. As written, the new test source does not compile, so the added regression coverage never runs.

Suggestion: Call an existing API entry point instead of a nonexistent repository method, or add the missing repository wrapper before using it here.

app/src/test/java/com/abk/kernel/data/repository/GitHubRepositoryNetworkTest.kt
```diff
@@ -0,0 +1,120 @@
...
+                    }
+                    MockResponse()
+                        .setResponseCode(200)
+                        .setBody(body)
+                }
+                else -> MockResponse().setResponseCode(404)
+            }
+        }
+
+        val result = repository.downloadReleaseAssetText(
```
xingguangcuicanbot, line 108: Location: app/src/test/java/com/abk/kernel/data/repository/GitHubRepositoryNetworkTest.kt:108

Issue: `downloadReleaseAssetText` is a `suspend` function, but this JUnit test is a regular synchronous `@Test` and invokes it directly. Kotlin will reject this at compile time, so the test suite fails to build until the call is wrapped in `runTest`/`runBlocking` (and the same pattern is applied anywhere else a suspend API is exercised).

Suggestion: Convert this test to a coroutine test, for example by using `runTest { ... }`, before calling the suspend repository method.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The production fix looks directionally correct, but the newly added network tests currently do not compile: one calls a nonexistent repository method, and another invokes a suspend API from a non-coroutine test.

### Decision
request_changes

### Notes

PR #181, opened 2026-07-26
Harden artifact verification: path traversal, unbounded reads, run-id binding

baizishiroko:
Three fixes on the download → verify → flash path, plus unit tests. Each one closes a gap where the Kotlin app is weaker than the Python CLI doing the equivalent job.

## 1. Path traversal in the downloaded artifact filename

`DownloadUtils.downloadArtifact()` built the destination as `"${artifact.name}.zip"` in the non-bundled branch. `artifact.name` comes from the GitHub API for whatever repo the app is pointed at.

Every other use of it in that file already goes through `safeFileName()` — lines 311, 315, 316, 388, 814, 858, 878, 896. This one line was the exception, so a name containing traversal segments wrote outside the downloads root. The consistency everywhere else is what makes this look like an oversight rather than a deliberate choice.

Safe to change: the zip is deleted right after `unzip()`, and the helper used for matching, `artifactStorageFolderName()`, is itself `safeFileName()` — so the sanitized name is what the rest of the path already expects.

## 2. Unbounded reads in the Kotlin verifier

`ArtifactVerification` read the manifest, signature, and payload with `readBytes()` and no size limit anywhere. The CLI verifier in `cli/abk.py` has enforced `MAX_MANIFEST_SIZE` / `MAX_SIGNATURE_SIZE` / `MAX_PAYLOAD_SIZE` all along (`cli/abk.py:103-105`).

Two things made this worse than a generic OOM:

- The manifest and signature are read **before** `verifyManifestSignature()`, so the allocation was reachable without any valid key.
- The payload size check ran *after* the array was already materialized, so it protected nothing.

This ports the CLI limits over, bounds each read by the running byte total rather than trusting the declared entry size, and digests the payload in chunks so it never has to fit in memory at once.

`readBundleManifest()` had the same unbounded read and is called before verification in two places (`prepareDownloadedArtifact`, and the flash security prompt), so it is bounded too.

## 3. Signed bundles were never bound to their run

`manifest.run_id` was deserialized and then never compared to anything — `verifyBundleFile()` took no expected-run-id parameter at all. The CLI does exactly this check (`cli/abk.py:2632`).

Because one fork signing key signs every build, a correctly signed bundle from an older run stayed valid indefinitely. Anyone able to substitute the artifact response could serve a previous, legitimately signed kernel — including one built before a security fix — and the app would accept it.

Adds an optional `expectedRunId` and threads the producing run id in at the call sites that know it. **Enforcement is skipped for ids `<= 0`**, so the unknown (`-1`) and prebuilt GKI (`-2`) sentinels behave exactly as before — there is a test pinning that.

## Also included

The CLI's `_safe_zip_member_name` check is ported to the manifest's `payload_name`, since `prepareDownloadedArtifact` builds a file path from it (`File(extractDir, verification.manifest.payloadName)`). Flagging it explicitly as a fourth change rather than folding it in silently — it is small and mirrors the CLI, but it was not one of the three reported issues.

## Not changed

`inspectBundleVerification()` has no run id available from its caller (`persistBundledDownloadEntry`), so that listing path stays unbound and keeps its current behaviour. Adding a parameter that is always null would have been dead plumbing.

## Tests

New `ArtifactVerificationTest` covers: happy path, run-id mismatch rejected, matching run id accepted, sentinel run ids not gating, oversized manifest rejected, oversized signature rejected, tampered payload rejected, and traversal/absolute names rejected.

## Verification status

The three fixes and the tests were written against the code as read, but **I could not compile or run them** — this environment has no JVM or Gradle, so `./gradlew test` has not been executed. Please let CI confirm before merging.

### DIFF
--- a/app/src/main/java/com/abk/kernel/utils/ArtifactVerification.kt
+++ b/app/src/main/java/com/abk/kernel/utils/ArtifactVerification.kt
@@ -3,12 +3,14 @@ package com.abk.kernel.utils
 import com.abk.kernel.data.model.ArtifactType
 import com.google.gson.Gson
 import com.google.gson.annotations.SerializedName
+import java.io.ByteArrayOutputStream
 import java.io.File
 import java.security.MessageDigest
 import java.security.PublicKey
 import java.security.Signature
 import java.util.Base64
 import java.util.Locale
+import java.util.zip.ZipEntry
 import java.util.zip.ZipFile
 
 data class SignedBundleManifest(
@@ -41,6 +43,16 @@ object ArtifactVerification {
     const val MANIFEST_FILE_NAME: String = "ABK_BUNDLE_MANIFEST.json"
     const val SIGNATURE_FILE_NAME: String = "ABK_BUNDLE_MANIFEST.sig"
 
+    // Mirrors the CLI limits in cli/abk.py so both verifiers refuse the same
+    // oversized input instead of allocating it. These bound the read itself:
+    // the manifest and signature are read before the signature is checked, so
+    // an unbounded read here is reachable without any valid key.
+    const val MAX_MANIFEST_SIZE: Long = 1L * 1024 * 1024
+    const val MAX_SIGNATURE_SIZE: Long = 64L * 1024
+    const val MAX_PAYLOAD_SIZE: Long = 8L * 1024 * 1024 * 1024
+
+    private const val COPY_BUFFER_SIZE = 64 * 1024
+
     private val gson = Gson()
 
     fun requiresTrustedBundle(type: ArtifactType): Boolean = when (type) {
@@ -52,10 +64,12 @@ object ArtifactVerification {
     fun readBundleManifest(bundleFile: File): SignedBundleManifest? = runCatching {
         ZipFile(bundleFile).use { zip ->
             zip.getEntry(MANIFEST_FILE_NAME)?.let { manifestEntry ->
-                gson.fromJson(
-                    zip.getInputStream(manifestEntry).use { String(it.readBytes(), Charsets.UTF_8) },
-                    SignedBundleManifest::class.java
-                )
+                readEntryAtMost(zip, manifestEntry, MAX_MANIFEST_SIZE)?.let { manifestBytes ->
+                    gson.fromJson(
+                        String(manifestBytes, Charsets.UTF_8),
+                        SignedBundleManifest::class.java
+                    )
+                }
             }
         }
     }.getOrNull()
@@ -63,7 +77,8 @@ object ArtifactVerification {
     fun verifyBundleFile(
         bundleFile: File,
         expectedType: ArtifactType? = null,
-        publicKeyPem: String? = null
+        publicKeyPem: String? = null,
+        expectedRunId: Long? = null
     ): BundleVerificationResult {
         if (!bundleFile.isFile || !bundleFile.name.lowercase(Locale.ROOT).endsWith(".bundle.zip")) {
             return failureFor(
@@ -96,8 +111,20 @@ object ArtifactVerification {
                         "Missing $SIGNATURE_FILE_NAME",
                         BundleVerificationFailureReason.MISSING_SIGNATURE
                     )
-                val manifestBytes = zip.getInputStream(manifestEntry).use { it.readBytes() }
-                val signatureBytes = zip.getInputStream(signatureEntry).use { it.readBytes() }
+                val manifestBytes = readEntryAtMost(zip, manifestEntry, MAX_MANIFEST_SIZE)
+                    ?: return failureFor(
+                        bundleFile.name,
+                        expectedType,
+                        "$MANIFEST_FILE_NAME exceeds $MAX_MANIFEST_SIZE bytes",
+                        BundleVerificationFailureReason.OTHER
+                    )
+                val signatureBytes = readEntryAtMost(zip, signatureEntry, MAX_SIGNATURE_SIZE)
+                    ?: return failureFor(
+                        bundleFile.name,
+                        expectedType,
+                        "$SIGNATURE_FILE_NAME exceeds $MAX_SIGNATURE_SIZE bytes",
+                        BundleVerificationFailureReason.OTHER
+                    )
                 val manifest = gson.fromJson(String(manifestBytes, Charsets.UTF_8), SignedBundleManifest::class.java)
                 if (!verifyManifestSignature(publicKey, manifestBytes, signatureBytes)) {
                     return BundleVerificationResult(
@@ -114,13 +141,26 @@ object ArtifactVerification {
                 if (manifest.bundleName != bundleFile.name) {
                     return BundleVerificationResult(manifest, false, "Bundle filename mismatch")
                 }
+                // A signed bundle stays valid indefinitely unless it is bound to the run
+                // that produced it, so an older signed build can otherwise be replayed in
+                // place of a newer one. Sentinel ids (unknown, prebuilt) stay exempt.
+                if (expectedRunId != null && expectedRunId > 0L && manifest.runId != expectedRunId) {
+                    return BundleVerificationResult(manifest, false, "Bundle run id mismatch")
+                }
+                if (!isSafeZipMemberName(manifest.payloadName)) {
+                    return BundleVerificationResult(manifest, false, "Unsafe payload name ${manifest.payloadName}")
+                }
+                if (manifest.payloadSizeBytes < 0L || manifest.payloadSizeBytes > MAX_PAYLOAD_SIZE) {
+                    return BundleVerificationResult(manifest, false, "Payload size out of range")
+                }
                 val payloadEntry = zip.getEntry(manifest.payloadName)
                     ?: return BundleVerificationResult(manifest, false, "Missing payload entry ${manifest.payloadName}")
-                val payloadBytes = zip.getInputStream(payloadEntry).use { it.readBytes() }
-                if (payloadBytes.size.toLong() != manifest.payloadSizeBytes) {
+                val payload = digestEntry(zip, payloadEntry, MAX_PAYLOAD_SIZE)
+                    ?: return BundleVerificationResult(manifest, false, "Payload exceeds $MAX_PAYLOAD_SIZE bytes")
+                if (payload.size != manifest.payloadSizeBytes) {
                     return BundleVerificationResult(manifest, false, "Payload size mismatch")
                 }
-                if (sha256(payloadBytes) != normalizeDigest(manifest.payloadSha256)) {
+                if (payload.digest != normalizeDigest(manifest.payloadSha256)) {
                     return BundleVerificationResult(manifest, false, "Payload digest mismatch")
                 }
                 BundleVerificationResult(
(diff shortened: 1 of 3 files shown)


xingguangcuican6666:
huh？

Mazha0309 (changes requested):
Reviewed the diff at 47c437b against the current `dev` codebase. The three hardening fixes address real issues and the overall direction is right. What I verified:

- **Fix 1 (path traversal)** is consistent and safe: `safeFileName()` strips everything outside `[A-Za-z0-9._ -]` (DownloadUtils.kt:939), so `../../evil` becomes `.._.._evil`; the zip is deleted right after `unzip()` (line 393) and the surviving paths use the same sanitized name.
- **Fix 2 (unbounded reads)** matches the CLI limits exactly (cli/abk.py:103-105), bounds every read by the running byte total instead of trusting the declared entry size, and digests the payload in chunks so it never has to fit in memory. The new checks run **after** signature verification, which is the correct order.
- **Fix 3 (run-id binding)** is semantically sound: the manifest `run_id` is `GITHUB_RUN_ID` of the build run (build.yml:5686), and `DownloadedArtifact.runId` is stored as `run?.id` at download time (DownloadUtils.kt:433), so in the primary download flow both sides come from the same run. Skipping enforcement for ids `<= 0` (unknown `-1`, prebuilt `-2`) is sensible and tested.

Issues to address before this is mergeable:

1. **False-rejection risk for mirrored artifacts — please confirm.** `mirror-custom-artifacts.yml` and `release-existing-full-feature-run.yml` download artifacts from a source run and re-upload them byte-for-byte to a *different* run. The bundle keeps the source `run_id`, but when a user browses the mirror run in the app, `artifact.runId` is the mirror run id, so `expectedRunId != manifest.runId` and a correctly signed, legitimate bundle is rejected with "Bundle run id mismatch". The CLI doesn't have this problem because the expected run id is passed explicitly by the caller. If the app ever surfaces artifacts from these mirror/release runs (forks do use these workflows), this check needs a way to distinguish them, or at minimum this scenario should be confirmed as out of scope.
2. **The PR description says `inspectBundleVerification()` has "no run id available from its caller" — that is inaccurate.** The chain is `inspectBundleVerification` <- `persistBundledDownloadEntry` <- `createBundledDownloadEntries`, and both call sites have a run id in scope: `downloadArtifact()` has `run: WorkflowRun?` (line 377) and `downloadDirectAsset()` has `runId` (line 621). The flash-time checks (`prepareDownloadedArtifact`, `precheckFlashSecurity`) do bind the run id via `DownloadedArtifact.runId`, so the security property still holds end-to-end; only the download-list `verified` badge in the notice-bundle path stays unbound. That trade-off is acceptable, but the stated reason is wrong and a reviewer will flag it.
3. **One CLI check was not ported:** `cli/abk.py:2623` rejects a `payload_name` that equals the manifest or signature entry names (`payload_name not in {manifest_name, signature_name}`). Not an exploitable gap since the manifest is authenticated, but it contradicts the "mirrors the CLI" claim.
4. Minor: in the high-risk fallback path (`prepareDownloadedArtifact`, DownloadUtils.kt:858-864) `payloadName` from `readBundleManifest()` builds `File(extractDir, payloadName)` without the new `isSafeZipMemberName` check. `unzip()`'s canonical-path guard (line 1024) already blocks extraction traversal and this is an explicitly user-consented path, so it is acceptable — but a comment acknowledging that would be good.

I could not compile or run the new tests (no local JVM/Android toolchain), so CI must confirm they pass before merging.

app/src/main/java/com/abk/kernel/utils/ArtifactVerification.kt
```diff
@@ -114,13 +141,26 @@ object ArtifactVerification {
                 if (manifest.bundleName != bundleFile.name) {
                     return BundleVerificationResult(manifest, false, "Bundle filename mismatch")
                 }
+                // A signed bundle stays valid indefinitely unless it is bound to the run
+                // that produced it, so an older signed build can otherwise be replayed in
+                // place of a newer one. Sentinel ids (unknown, prebuilt) stay exempt.
+                if (expectedRunId != null && expectedRunId > 0L && manifest.runId != expectedRunId) {
```
Mazha0309, line 147: Replay protection is correct for the primary flow (manifest `run_id` = `GITHUB_RUN_ID` of the build run, `DownloadedArtifact.runId` = the run browsed at download time, and this check runs after signature verification — good order). One scenario to confirm before merge: `mirror-custom-artifacts.yml` and `release-existing-full-feature-run.yml` re-upload artifacts byte-for-byte from a source run to a *new* run, so a user browsing the mirror run would see `expectedRunId` (mirror run id) != `manifest.runId` (source run id) and a legitimately signed bundle would be rejected with "Bundle run id mismatch". If those mirrored runs are surfaced in the app, this needs a fallback; otherwise please state they are out of scope.

app/src/main/java/com/abk/kernel/utils/DownloadUtils.kt
```diff
@@ -401,7 +401,8 @@ object DownloadUtils {
                         ArtifactVerification.verifyBundleFile(
                             candidate,
                             type,
-                            ForkSigningManager.publicKeyPemFromStoredValue(signingPublicKey)
+                            ForkSigningManager.publicKeyPemFromStoredValue(signingPublicKey),
+                            expectedRunId = run?.id
```
Mazha0309, line 405: This binds the direct path, but the notice-bundle path (`bundleWithNotices = true`, the default in MainViewModel) verifies via `createBundledDownloadEntries` -> `persistBundledDownloadEntry` -> `inspectBundleVerification`, which stays unbound. The PR description justifies this with "no run id available from its caller", but that is inaccurate: `run: WorkflowRun?` is in scope in `downloadArtifact()` (line 377) and `runId` in `downloadDirectAsset()` (line 621). Flash-time checks still bind through `DownloadedArtifact.runId`, so the security property holds end-to-end — only the download-list `verified` badge stays unbound. That trade-off is fine, but the stated rationale should be corrected.

Mazha0309 (commented):
Inline comments for the four points in my review above (the main review body stands).

app/src/main/java/com/abk/kernel/utils/ArtifactVerification.kt
```diff
@@ -114,13 +141,26 @@ object ArtifactVerification {
...
+                    return BundleVerificationResult(manifest, false, "Bundle run id mismatch")
+                }
+                if (!isSafeZipMemberName(manifest.payloadName)) {
```
Mazha0309, line 150: The CLI additionally rejects a `payload_name` equal to the manifest/signature entry names (`cli/abk.py:2623`: `payload_name not in {manifest_name, signature_name}`). Not exploitable here since the manifest is authenticated before this point, but if the intent is to mirror the CLI's `_safe_zip_member_name` behaviour, consider porting that part too.

app/src/main/java/com/abk/kernel/utils/DownloadUtils.kt
```diff
@@ -850,7 +852,8 @@ object DownloadUtils {
                 ArtifactVerification.verifyBundleFile(
                     source,
                     effectiveType,
-                    ForkSigningManager.publicKeyPemFromStoredValue(signingPublicKey)
+                    ForkSigningManager.publicKeyPemFromStoredValue(signingPublicKey),
+                    expectedRunId = artifact.runId
```
Mazha0309, line 856: Related consistency note: in the `allowHighRiskFallback` branch of this function, `payloadName` from `readBundleManifest()` still builds `File(extractDir, payloadName)` unchecked (~line 862). `unzip()`'s canonical-path guard already blocks extraction traversal and this is an explicitly user-consented path, so it is acceptable — but applying `isSafeZipMemberName` there too would keep the property uniform.

xingguangcuican6666:
huh？

app/src/main/java/com/abk/kernel/utils/ArtifactVerification.kt
```diff
@@ -52,18 +64,21 @@ object ArtifactVerification {
     fun readBundleManifest(bundleFile: File): SignedBundleManifest? = runCatching {
         ZipFile(bundleFile).use { zip ->
             zip.getEntry(MANIFEST_FILE_NAME)?.let { manifestEntry ->
-                gson.fromJson(
-                    zip.getInputStream(manifestEntry).use { String(it.readBytes(), Charsets.UTF_8) },
-                    SignedBundleManifest::class.java
```
xingguangcuicanbot, line 57: Location: app/src/main/java/com/abk/kernel/utils/ArtifactVerification.kt:67

Issue: The new payload-name validation only runs inside `verifyBundleFile`, but `prepareDownloadedArtifact` still consumes `readBundleManifest().payloadName` directly when verification is disabled. An untrusted manifest can therefore provide a traversal name, and `File(extractDir, payloadName)` may resolve outside the extraction directory to an existing app file.

Suggestion: Validate the manifest payload name at every call site that uses `readBundleManifest().payloadName`, including the verification-disabled path.

app/src/main/java/com/abk/kernel/utils/ArtifactVerification.kt
```diff
@@ -114,13 +141,26 @@ object ArtifactVerification {
                 if (manifest.bundleName != bundleFile.name) {
                     return BundleVerificationResult(manifest, false, "Bundle filename mismatch")
                 }
+                // A signed bundle stays valid indefinitely unless it is bound to the run
+                // that produced it, so an older signed build can otherwise be replayed in
```
xingguangcuicanbot, line 145: Location: app/src/main/java/com/abk/kernel/utils/ArtifactVerification.kt:150

Issue: A run-id mismatch or other verification failure can enter `allowHighRiskFallback`, where `verification.manifest.payloadName` is used without rechecking the new safety predicate. This allows fallback to bypass the payload traversal protection and select a file outside `extractDir`.

Suggestion: Require a safe payload name before using `verification.manifest.payloadName` in fallback, and apply the same validation to the legacy-manifest fallback.

app/src/main/java/com/abk/kernel/utils/ArtifactVerification.kt
```diff
@@ -114,13 +141,26 @@ object ArtifactVerification {
...
+                // place of a newer one. Sentinel ids (unknown, prebuilt) stay exempt.
+                if (expectedRunId != null && expectedRunId > 0L && manifest.runId != expectedRunId) {
+                    return BundleVerificationResult(manifest, false, "Bundle run id mismatch")
+                }
+                if (!isSafeZipMemberName(manifest.payloadName)) {
+                    return BundleVerificationResult(manifest, false, "Unsafe payload name ${manifest.payloadName}")
```
xingguangcuicanbot, line 151: Location: app/src/main/java/com/abk/kernel/utils/ArtifactVerification.kt:156

Issue: The verifier hashes one `ZipFile.getEntry(manifest.payloadName)`, while extraction later iterates the archive with `ZipInputStream` and can overwrite duplicate names. An archive with duplicate payload entries can therefore pass verification for one entry but flash different bytes. The CLI requires the payload name to occur exactly once.

Suggestion: Require exactly one non-directory payload entry before digesting it, matching the later extraction behavior.

```suggestion
val payloadEntries = zip.entries().asSequence()
    .filter { it.name == manifest.payloadName }
    .toList()
if (payloadEntries.size != 1 || payloadEntries[0].isDirectory) {
    return BundleVerificationResult(manifest, false, "Payload entry must occur exactly once")
}
val payloadEntry = payloadEntries.single()
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The hardening is incomplete: payload-name validation is bypassable in fallback and verification-disabled flows, and duplicate ZIP payload entries can make verified bytes differ from bytes later extracted and flashed.

### Decision
request_changes

### Notes

PR #186, opened 2026-08-01
feat(app): add frosted-glass blur to top/bottom bars

fanziyun:
## 概述

把 `miuix-blur` 的毛玻璃能力接入 ABK 的 Material 3 Expressive UI：顶栏、底栏和主要设置/内容页都改为可在滚动内容上方显示 frosted glass 效果，并保留了自定义背景参与 blur 的能力。

## 这次改了什么

- 新增 `app/src/main/java/com/abk/kernel/ui/blur/` 下的 blur 基础设施：`BlurState`、`BlurBackdrop`、`BlurHost`、`BlurScreenScaffold`、`BlurBackgroundPainter`
- 主界面与各功能页迁移到新的共享 scaffold，顶栏改为悬浮绘制，body 统一预留顶部空间
- `MainActivity` 的底部导航栏/导航轨也接入 blur
- 设置页新增 blur 总开关与“将自定义背景渲染到模糊”子开关，并保留界面不透明度的实时预览
- 独立的扩展管理页也接入同一套 blur / 背景配置
- 补齐 `app_dev` 的 blur / splashscreen 依赖，方便后续合并到 dev 分支
- 更新相关测试，确保仓库层网络测试继续覆盖 repository 行为

## 验证

- `gradle --no-daemon :app:compileDebugKotlin`
- `gradle --no-daemon :app:assembleDebug`
- CI 上的 `Unit Tests` 仍有 7 个与本 PR 无关的基线失败


app/src/main/java/com/abk/kernel/extensions/AbkExtensionManagerActivity.kt
```diff
@@ -269,7 +273,13 @@ fun AbkExtensionManagerScreen(
         return
     }
```
xingguangcuicanbot, line 283: Location: app/src/main/java/com/abk/kernel/extensions/AbkExtensionManagerActivity.kt:277

Issue: This `BlurConfig` hard-codes `backgroundUri = null` and `backgroundImageEnabled = false`, so the new "Render custom background into blur" option can never take effect on the extension-manager screen. Even when the page is visibly showing a custom background, the blur source is built as if no background exists.

Suggestion: Thread the current background URI and background-enabled state into `AbkExtensionManagerScreen` and use those values when constructing `BlurConfig`.

app/src/test/java/com/abk/kernel/data/repository/GitHubRepositoryNetworkTest.kt
```diff
@@ -53,9 +55,9 @@ class GitHubRepositoryNetworkTest {
                 )
         )
```
xingguangcuicanbot, line 57: Location: app/src/test/java/com/abk/kernel/data/repository/GitHubRepositoryNetworkTest.kt:58

Issue: The test name still says it verifies `GitHubRepository.getRepo`, but the assertion now calls `apiService.getRepo(...)` directly. That means the repository layer is no longer exercised here, so regressions in `GitHubRepository.getRepo`'s wrapping/mapping behavior would pass unnoticed.

Suggestion: Call `repository.getRepo(...)` in this test, not the raw Retrofit service, so the test continues to cover the repository contract.

```suggestion
val result = runBlocking { repository.getRepo("owner", "repo") }

assertTrue(result is Result.Success)
```

xingguangcuicanbot (dismissed):
## Automated PR Review

### Summary
The blur migration is broadly consistent, but the standalone extension-manager flow does not actually honor the new blur settings, and one updated network test no longer covers the repository layer it is supposed to validate.

### Decision
request_changes

### Notes

- app/src/main/java/com/abk/kernel/extensions/AbkExtensionManagerActivity.kt:108 `AbkExtensionManagerActivity` now wraps the screen in the new blur-capable UI, but it never reads `blurEnabled` / `blurBackgroundExpEnabled` from `PreferencesRepository`. Because `AbkExtensionManagerScreen` still defaults both flags to `false`, the standalone extension-manager activity ignores the user's blur setting entirely and always renders opaque bars.

app/src/main/java/com/abk/kernel/MainActivity.kt
```diff
@@ -561,37 +576,101 @@ private fun AbkMainScaffold(
...
+                        .fillMaxHeight()
+                        .zIndex(if (childPageVisible) 0f else 2f)
+                        .graphicsLayer {
+                            val hidden = 1f - navProgress
+                            translationX = -hidden * railHideDistancePx
+                            alpha = 1f - (hidden * 0.15f)
+                        },
+                    contentAlignment = Alignment.Center
+                ) {
+                    NavigationRail(
```
xingguangcuicanbot, line 633: Location: app/src/main/java/com/abk/kernel/MainActivity.kt:615

Issue: `blurActive` is computed specifically to avoid using the Miuix blur path when runtime shaders are unavailable, but both the tablet rail and phone bottom bar still apply `Modifier.blurEffect()` unconditionally. On Android 12/12L this leaves `textureBlur()` on the code path even though the bars have already fallen back to opaque colors, which is inconsistent with the PR's own compatibility gating and can still trigger unsupported rendering behavior on those devices.

Suggestion: Only attach `blurEffect()` when `blurActive` is true, for both `NavigationRail` and `NavigationBar`, so API 31-32 fully use the opaque fallback path instead of still invoking the blur modifier.

app/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools">
+
+    <uses-sdk tools:overrideLibrary="top.yukonga.miuix.kmp.blur" />
```
xingguangcuicanbot, line 5: Location: app/src/main/AndroidManifest.xml:5

Issue: `tools:overrideLibrary` disables the blur library's declared `minSdk` check for the entire app. That means API 26-30 installs now depend on every current and future initialization path inside `miuix-blur-android` remaining manually guarded; a single unguarded API 31+ reference in the library can crash the app before your own runtime checks execute. This is a fragile compatibility contract for a core dependency and should not be papered over at manifest-merge time.

Suggestion: Remove the manifest-level override and use an artifact/source-set arrangement that genuinely supports the app's `minSdk`, or otherwise isolate the dependency so pre-31 devices never load code from the blur library.

xingguangcuicanbot (dismissed):
## Automated PR Review

### Summary
The PR introduces two compatibility risks: the new blur path is still invoked by the main navigation bars on unsupported Android versions, and the manifest now force-bypasses the blur library's declared `minSdk`, making older-device startup dependent on library internals staying perfectly guarded.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/MainActivity.kt
fanziyun: Fixed in 2bbf8469: both NavigationRail and NavigationBar now attach `Modifier.blurEffect()` only when `blurActive` (enableBlur && backdrop != null && isRuntimeShaderSupported). On API 31-32 the modifier is not attached at all, so textureBlur never enters the composition tree — consistent with ExpressiveTopBar and the opaque fallback color.

app/src/main/AndroidManifest.xml
fanziyun: Kept deliberately. The manifest override only affects the manifest-merge minSdk check (no runtime effect); every miuix blur entry point is guarded before touching the library: rememberBlurBackdrop returns null when !isRenderEffectSupported() (API<31), and after 2bbf8469 Modifier.blurEffect() is only attached when isRuntimeShaderSupported() (API>=33). So below API 31 the library blur code is never invoked (and may not even load — lazy class loading), and below API 33 the textureBlur path is never in the composition. This matches the ReSukiSU reference (miuix-blur 0.9.2 + minSdk 26 + overrideLibrary). Raising minSdk to 33 would drop Android 8-12 support, which is what the guarded fallback exists to avoid.

app/src/main/java/com/abk/kernel/ui/blur/BlurBackdrop.kt
```diff
@@ -0,0 +1,105 @@
...
+ * @param backgroundDim Alpha of the dimmed base drawn under [backgroundPainter].
+ */
+@Composable
+fun rememberBlurBackdrop(
+    enableBlur: Boolean,
+    surfaceColor: Color,
+    backgroundPainter: Painter? = null,
+    backgroundDim: Float = AbkBlurBackgroundDim,
+): LayerBackdrop? {
+    if (!enableBlur || !isRenderEffectSupported()) return null
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/ui/blur/BlurBackdrop.kt:46

Issue: `rememberBlurBackdrop()` still creates a `LayerBackdrop` on Android 12 / 12L because it only checks `isRenderEffectSupported()`. In this PR the actual frosted effect is gated on `isRuntimeShaderSupported()` (API 33+), so API 31-32 devices fall back to opaque bars but still pay for full-screen backdrop capture and redraw on every screen, plus a second backdrop for the bottom nav. That is a real regression in rendering cost with no visible benefit.

Suggestion: Gate backdrop creation on the same runtime-shader capability that the bar blur uses, so unsupported devices skip the expensive capture path entirely.

```suggestion
if (!enableBlur || !isRenderEffectSupported() || !top.yukonga.miuix.kmp.shader.isRuntimeShaderSupported()) return null
```

xingguangcuicanbot (dismissed):
## Automated PR Review

### Summary
The blur fallback path still allocates and captures backdrops on API 31-32 even though no blur can render there. That should be fixed before merging.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/screens/SettingsScreen.kt
```diff
@@ -1773,20 +1801,25 @@ private fun defaultAppProfileTemplateJson(): String =
 
 @Composable
 private fun ThemeSettingsScreen(
-    padding: PaddingValues,
```
xingguangcuicanbot, line 1776: Location: app/src/main/java/com/abk/kernel/ui/screens/SettingsScreen.kt:1851

Issue: The transient "Blur enabled" confirmation is only cleared by a timer, so if the user turns blur on and then immediately turns it back off, the success row continues to render for up to 1.8 seconds even though the feature is now disabled. That leaves the settings UI in a contradictory state and is easy to hit when testing the new toggle.

Suggestion: Clear the feedback state as soon as blur is disabled, instead of waiting for the timeout.

```suggestion
fun onBlurSwitchChanged(enabled: Boolean) {
    onBlurEnabledChange(enabled)
    blurFeedbackVisible = enabled
}
LaunchedEffect(blurEnabled, blurFeedbackVisible) {
    if (blurEnabled && blurFeedbackVisible) {
        delay(1800)
        blurFeedbackVisible = false
    }
}
```

xingguangcuicanbot (dismissed):
## Automated PR Review

### Summary
The blur settings UI has one actionable state-management bug: the temporary "Blur enabled" confirmation can remain visible after the user has already disabled blur again.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/blur/BlurScreenScaffold.kt
```diff
@@ -0,0 +1,86 @@
...
+                subcompose("abk-top-bar", topBar)
+                    .map { it.measure(constraints.copy(minHeight = 0)) }
+            } else {
+                emptyList()
+            }
+            val barHeightPx = barPlaceables.maxOfOrNull { it.height } ?: 0
+            // The body is the backdrop's blur source; it must never include the top bar
+            // (that would sample the bar into itself and crash the render thread).
+            val contentPlaceable = subcompose("abk-body") {
+                Box(Modifier.fillMaxSize().blurSourceBody()) {
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/ui/blur/BlurScreenScaffold.kt:76

Issue: `BlurScreenScaffold` replaced Material3 `Scaffold`, but it no longer applies any safe-drawing/content insets to the body. Before this PR, every migrated screen received `PaddingValues` from `Scaffold`, which kept the last list items and controls above the gesture/navigation area. Now the body is laid out edge-to-edge and each screen only re-adds a top spacer, so pages like `SettingsScreen`, `ModuleRepositoryScreen`, `RootAuthorizationScreen`, and the child overlays can render their bottom content under the system bar on gesture-navigation devices.

Suggestion: Restore bottom safe-area handling in the shared scaffold, or pass equivalent inset padding back to call sites instead of only exposing `topBarHeight`.

app/src/main/AndroidManifest.xml
xingguangcuicanbot, line 5: Location: app/src/main/AndroidManifest.xml:5

Issue: This manifest change suppresses the blur library's declared `minSdk` check for the entire app. That means API 26-32 devices will now install code from `top.yukonga.miuix.kmp.blur` even if a future library update starts touching newer platform classes or resources before your runtime guards run. In that case the app can fail during class verification/resource inflation on launch, and the manifest override will have hidden the compatibility contract that would normally block the bad install.

Suggestion: Avoid `tools:overrideLibrary` here; use a dependency/build setup whose manifest requirements already match the app's supported API range, or isolate the blur implementation behind an API-specific variant.

app/src/main/java/com/abk/kernel/ui/screens/SettingsScreen.kt
```diff
@@ -1838,6 +1870,31 @@ private fun ThemeSettingsScreen(
...
+                onCheckedChange = onBlurEnabledChange
+            )
+            // The nested "render custom background into blur" item expands out from
+            // below the toggle when blur is enabled.
+            AnimatedVisibility(
+                visible = blurEnabled,
+                enter = fadeIn() + expandVertically(),
+                exit = fadeOut() + shrinkVertically()
+            ) {
+                ExpressiveSwitchItem(
```
xingguangcuicanbot, line 1888: Location: app/src/main/java/com/abk/kernel/ui/screens/SettingsScreen.kt:1995

Issue: `latestAlpha` is remembered only once, so it stops tracking external `alpha` updates after the first composition. If the persisted opacity changes outside an active drag (for example after reopening the page or after any future reset path), `onValueChangeFinished` can write back a stale value even though the slider is displaying the new `alpha`. The local state needs to be keyed or synchronized to the incoming prop.

Suggestion: Reinitialize the cached drag value when the `alpha` prop changes.

```suggestion
var latestAlpha by remember(alpha) { mutableFloatStateOf(alpha.coerceIn(0f, 1f)) }
```

xingguangcuicanbot (dismissed):
## Automated PR Review

### Summary
The PR introduces a solid blur abstraction, but I found three actionable regressions: the new shared scaffold drops bottom safe-area handling for migrated screens, the manifest now bypasses the blur library's minSdk contract for all devices, and the new opacity preview slider can persist a stale value because its local drag state is not synchronized with external updates.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/screens/SettingsScreen.kt
fanziyun: Resolved in 5e2b1ae1: the transient "blur enabled" confirmation row was removed entirely. Toggling blur now only expands/collapses the nested "render custom background into blur" item; there is no timer state to leave stale.

app/src/main/AndroidManifest.xml
fanziyun: Kept deliberately, and now strictly safe: a21821fe added `!isRuntimeShaderSupported()` to the rememberBlurBackdrop guard, so below API 33 no LayerBackdrop is created at all — the blur library code path is never reached (no backdrop, no blurSource capture, no textureBlur) below API 33, and the override only affects manifest merge (no runtime effect). This matches the ReSukiSU reference (miuix-blur 0.9.2 + minSdk 26 + overrideLibrary); raising minSdk to 33 would drop Android 8-12 support.

xingguangcuicanbot, line 5: Location: app/src/main/AndroidManifest.xml:5

Issue: `tools:overrideLibrary` suppresses the manifest merger's minSdk check for `top.yukonga.miuix.kmp.blur`, so this APK can now be installed on API levels the blur library itself may not support. If that dependency references newer framework classes during class loading, users on older devices will hit runtime crashes instead of being blocked at install/build time. Since the PR already contains explicit runtime fallbacks for unsupported blur devices, the safer fix is to use a blur artifact/version that is actually compatible with the app's `minSdk`, or isolate the dependency behind an API-specific source set, rather than forcing the manifest merge through.

Suggestion: Remove the `tools:overrideLibrary` override and make the dependency strategy match the app's supported API range. If the blur library really requires a higher minSdk, gate it at the build/dependency level instead of allowing unsupported installs.

xingguangcuicanbot (dismissed):
## Automated PR Review

### Summary
The PR introduces one blocking release-risk issue: it overrides the blur library's minSdk requirement in the manifest, which can allow unsupported devices to install an APK that may crash at runtime.

### Decision
request_changes

### Notes

xingguangcuican6666:
@codex review

app/src/main/java/com/abk/kernel/viewmodel/MainViewModel.kt
```diff
@@ -204,6 +205,8 @@ data class MainUiState(
     val customBackgroundUri: String? = null,
     val backgroundImageEnabled: Boolean = false,
     val uiSurfaceAlpha: Float = 1f,
+    val blurEnabled: Boolean = true,
+    val blurBackgroundExpEnabled: Boolean = false,
```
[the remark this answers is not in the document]
fanziyun: Confirmed and fixed in 2e9bf5e7: logout() now carries blurEnabled and blurBackgroundExpEnabled forward in the MainUiState rebuild, matching the other persisted prefs. Thanks for catching it.

PR #209, opened 2026-08-07
add(app): Add Android jailbreak mode

xingguangcuican6666:


app/src/main/java/com/abk/kernel/magica/AppZygotePreload.java
```diff
@@ -0,0 +1,87 @@
...
+import java.io.FileInputStream;
+import java.util.Properties;
+
+public class AppZygotePreload implements ZygotePreload {
+    private static final String TAG = "ABKMagica";
+    private static final String KSUD_LIBRARY_NAME = "libksud.so";
+    private static final String REQUEST_RELATIVE_PATH = "files/magica-jailbreak/request.properties";
+    private static final String REQUEST_MODULE_PATH = "modulePath";
+
+    private static native void forkDontCareAndExecKsud(
```
xingguangcuicanbot, line 19: Location: app/src/main/java/com/abk/kernel/magica/AppZygotePreload.java:19

Issue: `AppZygotePreload` is only referenced by the manifest's `android:zygotePreloadName` string, and its native method is resolved by the exact JNI symbol name. The release build enables R8 shrinking/obfuscation, but there is no keep rule for this class or method. R8 can remove or rename the preload class/method, causing the app-zygote preload to fail or the native lookup to throw in release APKs.

Suggestion: Add release keep rules for `com.abk.kernel.magica.AppZygotePreload` and its native method, or annotate the class/method with an appropriate keep annotation.

app/src/main/java/com/abk/kernel/magica/AppZygotePreload.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Override
+    public void doPreload(@NonNull ApplicationInfo appInfo) {
+        File ksud = new File(appInfo.nativeLibraryDir, KSUD_LIBRARY_NAME);
+        try {
+            if (!ksud.isFile()) {
+                throw new IllegalStateException("ksud does not exist: " + ksud);
+            }
+            Request request = readRequest(appInfo);
+            if (!request.delete()) {
```
xingguangcuicanbot, line 33: Location: app/src/main/java/com/abk/kernel/magica/AppZygotePreload.java:33

Issue: The request file is deleted before loading `abkksu` and before the fork/exec succeeds. Any transient failure in native-library loading, fork, `setuid`, or `exec` consumes the only request, so retrying jailbreak mode requires staging the module again and can leave the feature silently unavailable after a failed attempt.

Suggestion: Consume the request only after the native bootstrap has been successfully initiated, or preserve it when bootstrap fails so the operation can be retried.

```suggestion
            System.loadLibrary("abkksu");
            Log.d(TAG, "executing magica with external module: " + request.modulePath);
            forkDontCareAndExecKsud(ksud.getAbsolutePath(), appInfo.packageName, request.modulePath);
            if (!request.delete()) {
                Log.w(TAG, "failed to delete consumed request: " + request.file);
            }
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The release build does not keep the manifest-referenced app-zygote preload/native entry point, which can break the jailbreak path in minified APKs. The request is also consumed before bootstrap success, making failed attempts unnecessarily non-retryable.

### Decision
request_changes

### Notes

PR #217, opened 2026-08-10
fix(app): write patched boot images to the configured download directory

fanziyun:
## 背景

`patchAbkLkmBootImage`（ABK LKM boot/init_boot 修补流程）此前把产出的 patched `.img` 写到 app 私有外部目录 `Android/data/<pkg>/files/Download/abk-patched/`——文件管理器不可见、卸载即删，且与其它流程（下载/刷写走公共 `Download/ABK`）不一致。

## 改动

- 将 `MainUiState.downloadDirectory`（默认 `/storage/emulated/0/Download/ABK`）贯通 `RuntimeHomeScreen → AbkRootPatchScreen → patchAbkLkmBootImage`
- 输出目录改为三级优先链：公共 `downloadDirectory/abk-patched` → app 私有外部 `getExternalFilesDir(DOWNLOADS)/abk-patched` → 内部 `filesDir`，复用 `DownloadDirectoryUtils.normalizeDirectoryPath`，语义与 `DownloadUtils.resolveDownloadsRoot` 一致
- 仅在产出 patched 镜像的模式（非直接刷写）打印输出目录日志，避免 Direct Install / OTA 模式出现误导性提示

## 验收

- 修补后产物出现在 `Download/ABK/abk-patched/`，系统文件管理器可见
- 公共目录不可写时自动回退，无“所有文件访问”授权的用户不受影响
- root 路径以 app 进程 `canWrite()` 为准，保证最终 `outputImage.isFile` 校验与后续刷写能在 app 进程内执行
- `assembleDebug` 通过；完整单测 122 个中 8 个失败为 dev 既有基线（Robolectric 未 mock 等），与本改动无关

### DIFF
--- a/app/src/main/java/com/abk/kernel/utils/RootUtils.kt
+++ b/app/src/main/java/com/abk/kernel/utils/RootUtils.kt
@@ -476,7 +476,8 @@ object RootUtils {
         allowShell: Boolean = false,
         enableAdb: Boolean = false,
         localModulePath: String? = null,
-        onOutput: ((String) -> Unit)? = null
+        onOutput: ((String) -> Unit)? = null,
+        downloadDirectory: String? = null
     ): BootPatchResult {
         val sourceBoot = bootImagePath
             ?.takeIf { it.isNotBlank() }
@@ -511,10 +512,16 @@ object RootUtils {
                 stageBundledAbkLkmAsset(context, workDir, checkNotNull(asset))
             }
 
-            val outputDir = File(
-                context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS) ?: context.filesDir,
-                "abk-patched"
-            ).apply { mkdirs() }
+            val outputDir = resolvePatchOutputDir(downloadDirectory)
+                ?: File(
+                    context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS) ?: context.filesDir,
+                    "abk-patched"
+                ).apply { mkdirs() }
+            // Flash-only modes (Direct Install/OTA) patch the partition in place and don't hand
+            // the user a patched image, so the output-dir line would be noise there.
+            if (!flash) {
+                onOutput?.invoke(tr(R.string.ru_log_output_dir, outputDir.absolutePath))
+            }
             val moduleName = (asset?.let { "${it.variantId}-${it.kmi}" } ?: moduleFile.nameWithoutExtension)
                 .replace(Regex("""[^A-Za-z0-9._-]"""), "_")
             val outputName = "abk-${moduleName}-patched-${System.currentTimeMillis()}.img"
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/abk/kernel/utils/RootUtils.kt
```diff
@@ -511,10 +512,16 @@ object RootUtils {
                 stageBundledAbkLkmAsset(context, workDir, checkNotNull(asset))
             }
 
-            val outputDir = File(
-                context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS) ?: context.filesDir,
-                "abk-patched"
-            ).apply { mkdirs() }
+            val outputDir = resolvePatchOutputDir(downloadDirectory)
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/utils/RootUtils.kt:516

Issue: The new fallback chain still does not implement the advertised `public -> app-scoped external -> internal filesDir` behavior. `getExternalFilesDir(...)` is only null-checked here; if it returns a non-null path that cannot actually be created or written (for example, external storage is mounted read-only or the app-specific directory is unavailable), `outputDir` is left pointing at an unusable location and the patch flow fails instead of falling back to `filesDir`.

Suggestion: Validate the app-scoped external directory before using it, and only keep it if it is creatable and writable; otherwise continue to the `filesDir` fallback.

```suggestion
val outputDir = resolvePatchOutputDir(downloadDirectory)
    ?: sequenceOf(
        context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS)?.let { File(it, "abk-patched") },
        File(context.filesDir, "abk-patched")
    ).filterNotNull().firstOrNull { candidate ->
        (candidate.exists() || candidate.mkdirs()) && candidate.isDirectory && candidate.canWrite()
    } ?: return@runCatching BootPatchResult(
        false,
        listOf(tr(R.string.download_directory_create_failed, context.filesDir.absolutePath)),
        null
    )
```

app/src/main/java/com/abk/kernel/utils/RootUtils.kt
```diff
@@ -511,10 +512,16 @@ object RootUtils {
                 stageBundledAbkLkmAsset(context, workDir, checkNotNull(asset))
             }
 
-            val outputDir = File(
-                context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS) ?: context.filesDir,
-                "abk-patched"
-            ).apply { mkdirs() }
```
xingguangcuicanbot, line 517: Location: app/src/main/java/com/abk/kernel/utils/RootUtils.kt:515

Issue: `resolvePatchOutputDir(downloadDirectory)` now runs for every boot-patch invocation, including `--flash` / `--ota` modes. That means a direct install can create `Download/ABK/abk-patched` in public storage even though those flows do not present an output image to the user. Because the same `outputDir` is still passed into `buildBootPatchArgs`, any ksud implementation that leaves its `--out` artifact behind will silently drop patched images into the public downloads area during direct installs.

Suggestion: Only route output to the user-configured public download directory when `flash == false`. For direct-install / OTA flows, keep using app-private temporary storage so those modes do not create or leak public artifacts.

```suggestion
val outputDir = (if (!flash) resolvePatchOutputDir(downloadDirectory) else null)
    ?: sequenceOf(
        context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS)?.let { File(it, "abk-patched") },
        File(context.filesDir, "abk-patched")
    ).filterNotNull().firstOrNull { candidate ->
        (candidate.exists() || candidate.mkdirs()) && candidate.isDirectory && candidate.canWrite()
    } ?: return@runCatching BootPatchResult(
        false,
        listOf(tr(R.string.download_directory_create_failed, context.filesDir.absolutePath)),
        null
    )
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The patch mostly wires the configured download directory through correctly, but two correctness regressions remain in `RootUtils.patchAbkLkmBootImage`: the app-scoped-external-to-internal fallback chain is still incomplete, and flash-only modes now touch the public download directory even though they are not supposed to produce a user-facing patched image.

### Decision
request_changes

### Notes

xingguangcuicanbot (approved):
## Automated PR Review

### Summary
I reviewed the PR against `origin/dev...HEAD` and did not find any actionable issues in the changed diff. The new download-directory plumbing and fallback logic are consistent with the existing download path normalization and writable-directory checks, and the flash-only logging change does not appear to introduce a regression.

### Decision
approve

### Notes
Codex recommendation: approve

PR #219, opened 2026-08-11
feat(ui): Add Miuix blur and custom card backdrop blur support

fanziyun:
## Overview

This PR ports and integrates the `miuix-blur-android` library to add real-time frosted-glass blur effects to top bars and bottom navigation bars, along with software-based StackBlur support for custom background images on cards and tiles.

### 1. Miuix Blur Architecture (`app/src/main/java/com/abk/kernel/ui/blur/`)
- `BlurScreenScaffold` & `BlurHost`: Redesigned page scaffold layouts to place body content inside the capture backdrop while keeping top bars outside the capture region, preventing render thread crashes (SIGSEGV) caused by self-referential blur sampling.
- `blurSource` & `blurEffect`: Provided AGSL runtime shader blur (API 33+) for top bars and bottom navigation bars, automatically falling back to surface colors on unsupported devices.

### 2. Custom Background Card Blur (`BlurredCardBackdrop.kt`)
- `rememberBlurredCardBackground`: Downsamples custom background images and pre-blurs them on `Dispatchers.Default` using StackBlur when "Render custom background into blur" is enabled.
- `blurredCardBackground`: Allows cards and list items to sample their corresponding viewport rectangle for consistent background translucency.

### 3. Preferences & UI Performance (`SettingsScreen.kt` & `MainViewModel.kt`)
- `uiSurfaceAlphaPreview`: Keeps drag state local during surface alpha slider drag, updating memory state per frame and deferring DataStore disk writes until the drag finishes (`onValueChangeFinished`).
- Settings Controls: Added main "Blur" switch with expandable "Render custom background into blur" child toggle.

## Verification & Testing

- Unit Tests: Added `BlurConfigTest` to verify blur flag behavior.
- Compatibility: Verified AGSL shader rendering on API 33+ and fallback logic on older versions.
- Performance: Confirmed smooth slider dragging without I/O stuttering during surface opacity adjustments.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -42,7 +42,7 @@
             android:name=".MainActivity"
             android:exported="true"
             android:launchMode="singleTop"
-            android:windowSoftInputMode="adjustNothing"
+            android:windowSoftInputMode="adjustResize"
             android:theme="@style/Theme.ABK">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
@@ -91,7 +91,7 @@
         <activity
             android:name=".extensions.AbkExtensionManagerActivity"
             android:exported="false"
-            android:windowSoftInputMode="adjustNothing"
+            android:windowSoftInputMode="adjustResize"
             android:theme="@style/Theme.ABK" />
 
         <service
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -15,7 +15,14 @@
     <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
     <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
 
-    <!-- Runtime shader blur is guarded in code while the app still supports API 26+. -->
+    <!--
+      The miuix-blur AAR declares minSdk 33 while this app supports API 26+. The override
+      keeps the build working; safety is enforced at runtime. Every path into the library
+      (rememberBlurBackdrop, blurSource, blurEffect, blurSourceBody, isBlurActive) is gated
+      on isRuntimeShaderSupported() (API 33+), and the AAR registers no ContentProvider or
+      initializer, so nothing library-side executes on API 26-32. Keep these guards intact
+      if the dependency version or call sites change.
+    -->
     <uses-sdk tools:overrideLibrary="top.yukonga.miuix.kmp.blur" />
 
     <application
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -42,7 +42,7 @@
             android:name=".MainActivity"
             android:exported="true"
             android:launchMode="singleTop"
-            android:windowSoftInputMode="adjustResize"
+            android:windowSoftInputMode="adjustNothing"
             android:theme="@style/Theme.ABK">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
@@ -91,6 +91,7 @@
         <activity
             android:name=".extensions.AbkExtensionManagerActivity"
             android:exported="false"
+            android:windowSoftInputMode="adjustNothing"
             android:theme="@style/Theme.ABK" />
 
         <service
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
@@ -13,6 +14,9 @@
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
     <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
     <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
+
+    <!-- Runtime shader blur is guarded in code while the app still supports API 26+. -->
+    <uses-sdk tools:overrideLibrary="top.yukonga.miuix.kmp.blur" />
 
     <application
         android:name=".AbkApplication"
--- a/app/src/main/java/com/abk/kernel/ui/blur/BlurBackdrop.kt
+++ b/app/src/main/java/com/abk/kernel/ui/blur/BlurBackdrop.kt
@@ -1,6 +1,5 @@
 package com.abk.kernel.ui.blur
 
-import android.os.Build
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.geometry.Size
@@ -23,6 +22,13 @@ import kotlin.math.max
 
 internal const val AbkBlurRadius = 25f
 internal const val AbkBlurBackgroundDim = 0.35f
+
+/**
+ * Caps the surface tint applied over the frosted-glass backdrop. Keeping this below
+ * 1 keeps the blur visible even when [LocalUiSurfaceAlpha] is at its default 1f
+ * (no custom background, or background at 100% opacity).
+ */
+internal const val AbkBlurTintAlpha = 0.85f
 
 /** Creates a [LayerBackdrop] capturing content drawn beneath blurred bars. */
 @Composable
@@ -69,7 +75,9 @@ private fun ContentDrawScope.drawCroppedPainter(painter: Painter) {
 /** Marks content as the blur source for the active backdrop. */
 @Composable
 fun Modifier.blurSource(): Modifier {
-    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) return this
+    // layerBackdrop needs AGSL runtime shaders (API 33+). Gating on the real
+    // capability instead of a lower SDK constant keeps the fallback accurate.
+    if (!isBlurCapableDevice()) return this
     return LocalBlurState.current?.let { backdrop ->
         this.then(Modifier.layerBackdrop(backdrop))
     } ?: this
@@ -78,11 +86,14 @@ fun Modifier.blurSource(): Modifier {
 /** Applies a frosted-glass effect using the active backdrop. */
 @Composable
 fun Modifier.blurEffect(blendColor: Color = Color.Unspecified): Modifier {
-    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) return this
+    // textureBlur needs AGSL runtime shaders (API 33+).
+    if (!isBlurCapableDevice()) return this
     return LocalBlurState.current?.let { backdrop ->
         val effective = if (blendColor == Color.Unspecified) {
             MaterialTheme.colorScheme.surfaceContainer.copy(
-                alpha = LocalUiSurfaceAlpha.current.coerceIn(0f, 1f)
+                // Decoupled from LocalUiSurfaceAlpha so the frosted glass stays
+                // visible at the default opaque surface alpha (see AbkBlurTintAlpha).
+                alpha = (LocalUiSurfaceAlpha.current * AbkBlurTintAlpha).coerceIn(0f, 1f)
             )
         } else {
             blendColor
--- a/app/src/main/java/com/abk/kernel/ui/blur/BlurBackdrop.kt
+++ b/app/src/main/java/com/abk/kernel/ui/blur/BlurBackdrop.kt
@@ -24,15 +24,7 @@ import kotlin.math.max
 internal const val AbkBlurRadius = 25f
 internal const val AbkBlurBackgroundDim = 0.35f
 
-/**
- * Creates a [LayerBackdrop] capturing the content drawn beneath the blurred bars.
- *
- * The draw callback first paints an optional background [Painter] (the custom
- * background image, when enabled), then a dimmed surface base for contrast, then the
- * content that will be captured as the blur source. Returns `null` when blur is
- * disabled or the device cannot run the frosted-glass shader (API < 33), in which
- * case the app falls back to opaque surfaces.
- */
+/** Creates a [LayerBackdrop] capturing content drawn beneath blurred bars. */
 @Composable
 fun rememberBlurBackdrop(
     enableBlur: Boolean,
@@ -74,12 +66,7 @@ private fun ContentDrawScope.drawCroppedPainter(painter: Painter) {
     }
 }
 
-/**
- * Marks content as the blur source for the active backdrop.
- *
- * Attach to the page content box (inside the Scaffold body, not the bar itself) so
- * that content scrolling beneath the bars is captured and blurred.
- */
+/** Marks content as the blur source for the active backdrop. */
 @Composable
 fun Modifier.blurSource(): Modifier {
     if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) return this
@@ -88,16 +75,7 @@ fun Modifier.blurSource(): Modifier {
     } ?: this
 }
 
-/**
- * Applies a frosted-glass effect to a bar using the active backdrop.
- *
- * Attach to a top/bottom bar whose container color is [Color.Transparent] when the
- * backdrop is active, so the blurred content shows through.
- *
- * @param blendColor Optional tint blended over the blurred content. Defaults to the
- * theme's surfaceContainer using the live "界面不透明度" value directly, matching
- * the card tint and ReSukiSU's card-alpha behavior.
- */
+/** Applies a frosted-glass effect using the active backdrop. */
 @Composable
 fun Modifier.blurEffect(blendColor: Color = Color.Unspecified): Modifier {
     if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) return this
--- a/app/src/main/java/com/abk/kernel/ui/blur/BlurBackdrop.kt
+++ b/app/src/main/java/com/abk/kernel/ui/blur/BlurBackdrop.kt
@@ -1,0 +1,125 @@
+package com.abk.kernel.ui.blur
+
+import android.os.Build
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.geometry.Size
+import androidx.compose.ui.geometry.isSpecified
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.RectangleShape
+import androidx.compose.ui.graphics.drawscope.ContentDrawScope
+import androidx.compose.ui.graphics.drawscope.clipRect
+import androidx.compose.ui.graphics.drawscope.translate
+import androidx.compose.ui.graphics.painter.Painter
+import top.yukonga.miuix.kmp.blur.BlendColorEntry
+import top.yukonga.miuix.kmp.blur.BlurColors
+import top.yukonga.miuix.kmp.blur.LayerBackdrop
+import top.yukonga.miuix.kmp.blur.layerBackdrop
+import top.yukonga.miuix.kmp.blur.rememberLayerBackdrop
+import top.yukonga.miuix.kmp.blur.textureBlur
+import com.abk.kernel.ui.theme.LocalUiSurfaceAlpha
+import kotlin.math.max
+
+internal const val AbkBlurRadius = 25f
+internal const val AbkBlurBackgroundDim = 0.35f
+
+/**
+ * Creates a [LayerBackdrop] capturing the content drawn beneath the blurred bars.
+ *
+ * The draw callback first paints an optional background [Painter] (the custom
+ * background image, when enabled), then a dimmed surface base for contrast, then the
+ * content that will be captured as the blur source. Returns `null` when blur is
+ * disabled or the device cannot run the frosted-glass shader (API < 33), in which
+ * case the app falls back to opaque surfaces.
+ */
+@Composable
+fun rememberBlurBackdrop(
+    enableBlur: Boolean,
+    surfaceColor: Color,
+    backgroundPainter: Painter? = null,
+    backgroundDim: Float = AbkBlurBackgroundDim,
+): LayerBackdrop? {
+    if (!enableBlur || !isBlurCapableDevice()) return null
+    return rememberLayerBackdrop {
+        if (backgroundPainter != null) {
+            drawCroppedPainter(backgroundPainter)
+        } else {
+            drawRect(surfaceColor)
+        }
+        drawRect(surfaceColor.copy(alpha = backgroundDim))
+        drawContent()
+    }
+}
+
+private fun ContentDrawScope.drawCroppedPainter(painter: Painter) {
+    val targetSize = drawContext.size
+    val sourceSize = painter.intrinsicSize
+    if (!sourceSize.isSpecified || sourceSize.width <= 0f || sourceSize.height <= 0f) {
+        with(painter) { draw(size = targetSize) }
+        return
+    }
+
+    val scale = max(
+        targetSize.width / sourceSize.width,
+        targetSize.height / sourceSize.height,
+    )
+    val scaledSize = Size(sourceSize.width * scale, sourceSize.height * scale)
+    val left = (targetSize.width - scaledSize.width) / 2f
+    val top = (targetSize.height - scaledSize.height) / 2f
+    clipRect {
+        translate(left = left, top = top) {
+            with(painter) { draw(size = scaledSize) }
+        }
+    }
+}
+
+/**
+ * Marks content as the blur source for the active backdrop.
+ *
+ * Attach to the page content box (inside the Scaffold body, not the bar itself) so
+ * that content scrolling beneath the bars is captured and blurred.
+ */
+@Composable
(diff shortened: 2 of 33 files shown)


app/src/main/java/com/abk/kernel/ui/components/AppBackgroundHost.kt
```diff
@@ -35,10 +51,19 @@ fun AppBackgroundHost(
...
-            LocalUiSurfaceAlpha provides if (hasBackground) uiSurfaceAlpha.coerceIn(0f, 1f) else 1f,
-            LocalAppBackgroundEnabled provides hasBackground
+            LocalBlurBackgroundAnchor provides backgroundCoordinates,
         ) {
-            content()
+            val blurredCardBackground = rememberBlurredCardBackground(
+                uri = backgroundUri,
+                enabled = blurBackgroundEnabled && hasBackground,
+            )
+            CompositionLocalProvider(
```
xingguangcuicanbot, line 102: Location: app/src/main/java/com/abk/kernel/ui/components/AppBackgroundHost.kt:62

Issue: `LocalUiSurfaceAlpha` now remains below `1f` even when no custom background is active. Disabling or removing a background after lowering its opacity leaves app surfaces translucent over the solid background, changing the previous opaque fallback and potentially reducing contrast throughout the UI.

Suggestion: Preserve the previous opaque-surface behavior whenever no custom background is active.

```suggestion
                LocalUiSurfaceAlpha provides if (hasBackground) {
                    uiSurfaceAlpha.coerceIn(0f, 1f)
                } else {
                    1f
                },
```

app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt
```diff
@@ -0,0 +1,436 @@
...
+        return null
+    }
+    val context = LocalContext.current
+    var bitmap by remember(uri, widthPx, heightPx) { mutableStateOf<BlurredCardBackground?>(null) }
+    LaunchedEffect(uri, widthPx, heightPx) {
+        bitmap = withContext(Dispatchers.Default) {
+            runCatching {
+                val loader = context.imageLoader
+                val result = loader.execute(
+                    ImageRequest.Builder(context).data(uri).allowHardware(false).build()
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt:72

Issue: Coil decodes the complete source image without a size constraint before the software blur downsamples it. Large user-selected images can therefore allocate a full-resolution bitmap plus temporary blur buffers, causing substantial memory spikes or `OutOfMemoryError` when optional card blur is enabled.

Suggestion: Constrain Coil's decode size to the downsampled viewport before running StackBlur.

```suggestion
                    ImageRequest.Builder(context)
                        .data(uri)
                        .size(
                            (widthPx / AbkCardBlurDownsample).coerceAtLeast(1),
                            (heightPx / AbkCardBlurDownsample).coerceAtLeast(1),
                        )
                        .allowHardware(false)
                        .build()
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
Found two actionable issues: reduced surface alpha persists after disabling the custom background, and optional card blur decodes unbounded source images with OOM risk.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/components/AppBackgroundHost.kt
xingguangcuicanbot, line 102: Location: app/src/main/java/com/abk/kernel/ui/components/AppBackgroundHost.kt:62

Issue: This now applies the saved `uiSurfaceAlpha` even when no background image is enabled, so screens fall back to translucent surfaces instead of the previous opaque default. That makes the whole app unexpectedly transparent whenever the background is off.

Suggestion: Restore the previous opaque fallback when no background image is active.

```suggestion
                LocalUiSurfaceAlpha provides if (hasBackground) uiSurfaceAlpha.coerceIn(0f, 1f) else 1f,
                LocalAppBackgroundEnabled provides hasBackground,
```

app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt
```diff
@@ -0,0 +1,443 @@
...
+
+@Composable
+fun rememberBlurredCardBackground(
+    uri: String?,
+    enabled: Boolean,
+): BlurredCardBackground? {
+    val viewportSize = LocalBlurBackgroundAnchor.current?.size
+    val widthPx = viewportSize?.width ?: 0
+    val heightPx = viewportSize?.height ?: 0
+    if (!enabled || uri.isNullOrBlank() || !isBlurCapableDevice() || widthPx <= 0 || heightPx <= 0) {
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt:62

Issue: The software blur path is still blocked behind `isBlurCapableDevice()`, so the StackBlur fallback never renders on pre-API 33 devices. That defeats the custom card/tile blur feature on the very devices that need the fallback.

Suggestion: Remove the runtime-shader capability gate from the software blur path.

```suggestion
    if (!enabled || uri.isNullOrBlank() || widthPx <= 0 || heightPx <= 0) {
        return null
    }
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
Two regressions: opaque-surface fallback is broken when no background image is enabled, and the software card blur fallback is incorrectly disabled on non-shader devices.

### Decision
request_changes

### Notes

app/src/main/AndroidManifest.xml
```diff
@@ -14,6 +15,9 @@
     <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
     <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
 
+    <!-- Runtime shader blur is guarded in code while the app still supports API 26+. -->
```
xingguangcuicanbot: Location: app/src/main/AndroidManifest.xml:19

Issue: `tools:overrideLibrary` suppresses the blur AAR's declared `minSdk` check for every install target. The new app-side `Build.VERSION` guards only protect your own call sites; they do not stop class verification or any library-side initialization from running on API 26-32. If `miuix-blur-android` references newer platform APIs outside the guarded paths, this turns into a startup crash on older devices instead of a build-time failure.

Suggestion: Do not override the library's `minSdk` unless the entire AAR has been audited for pre-supported-API safety. A safer fix is to isolate the dependency behind a flavor/module that is only packaged on supported API levels, or replace it with code that does not require manifest-level `minSdk` bypassing.

app/src/main/java/com/abk/kernel/ui/blur/BlurBackdrop.kt
```diff
@@ -0,0 +1,103 @@
...
+    } ?: this
+}
+
+/** Applies a frosted-glass effect using the active backdrop. */
+@Composable
+fun Modifier.blurEffect(blendColor: Color = Color.Unspecified): Modifier {
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) return this
+    return LocalBlurState.current?.let { backdrop ->
+        val effective = if (blendColor == Color.Unspecified) {
+            MaterialTheme.colorScheme.surfaceContainer.copy(
```
xingguangcuicanbot, line 95: Location: app/src/main/java/com/abk/kernel/ui/blur/BlurBackdrop.kt:84

Issue: The new blur blend color uses `LocalUiSurfaceAlpha` as-is. `AppBackgroundHost` provides `1f` whenever no custom background is enabled, so the default top/bottom-bar blur path ends up blending an opaque `surfaceContainer` over the blurred content. In practice that makes the new Blur switch appear ineffective for the default configuration, because users only see the frosted effect after separately lowering UI opacity.

Suggestion: Decouple the blur tint alpha from `LocalUiSurfaceAlpha`, or cap the blend alpha below 1 so the blur remains visible even when surface opacity is left at its default value.

app/src/main/java/com/abk/kernel/ui/screens/SettingsScreen.kt
```diff
@@ -1957,20 +1986,25 @@ private fun defaultAppProfileTemplateJson(): String =
...
-    padding: PaddingValues,
+    topBarHeight: Dp,
     themeMode: String,
     dynamicColorEnabled: Boolean,
     customThemeColorArgb: Int?,
     customAccentColorArgb: Int?,
     backgroundUri: String?,
     backgroundImageEnabled: Boolean,
     uiSurfaceAlpha: Float,
+    blurEnabled: Boolean,
```
xingguangcuicanbot, line 1998: Location: app/src/main/java/com/abk/kernel/ui/screens/SettingsScreen.kt:2061

Issue: The entire Blur settings block is hidden behind `isBlurCapableDevice()`, but this PR also adds a software StackBlur path for cards/tiles that does not depend on runtime shaders. As written, API 26-32 devices can never enable the custom-background card blur feature described in the PR, because the toggle is unreachable and `wantsBackgroundPainter` is tied to the same master blur flag.

Suggestion: Expose the custom-background blur toggle independently from AGSL/runtime-shader support, or otherwise decouple the software card-blur path from the hardware bar-blur capability check.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The PR introduces a risky `minSdk` override for the new blur dependency, makes the default blur path effectively opaque unless UI opacity is lowered, and unintentionally disables the new software card-blur feature on pre-runtime-shader devices.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/components/AppBackgroundHost.kt
```diff
@@ -17,28 +29,57 @@ fun AppBackgroundHost(
...
     val hasBackground = backgroundEnabled && !backgroundUri.isNullOrBlank()
     val colorScheme = MaterialTheme.colorScheme
+    var backgroundCoordinates by remember { mutableStateOf<LayoutCoordinates?>(null) }
+    // One shared painter for the visible background image and every blur backdrop.
+    val backgroundPainter = if (hasBackground) {
+        rememberAsyncImagePainter(
+            model = backgroundUri,
+            contentScale = ContentScale.Crop,
+        )
+    } else {
```
xingguangcuicanbot, line 66: Location: app/src/main/java/com/abk/kernel/ui/components/AppBackgroundHost.kt:45

Issue: Switching the main wallpaper from `AsyncImage` to `rememberAsyncImagePainter` without a size resolver changes Coil's request behavior: this painter path loads the source at its original dimensions. Large user wallpapers will now be decoded at full resolution for every app start, which is a significant memory/CPU regression and can trigger OOMs on lower-end devices.

Suggestion: Keep the shared painter, but build it with a constraints-based size resolver (or another explicit screen-sized request) so the decode stays bounded to the viewport like the old `AsyncImage` path did.

app/src/main/java/com/abk/kernel/ui/components/AppBackgroundHost.kt
```diff
@@ -17,28 +29,57 @@ fun AppBackgroundHost(
...
+                painter = backgroundPainter,
                 contentDescription = null,
-                contentScale = ContentScale.Crop,
                 modifier = Modifier.fillMaxSize()
             )
         }
         CompositionLocalProvider(
-            LocalUiSurfaceAlpha provides if (hasBackground) uiSurfaceAlpha.coerceIn(0f, 1f) else 1f,
-            LocalAppBackgroundEnabled provides hasBackground
+            LocalBlurBackgroundAnchor provides backgroundCoordinates,
```
xingguangcuicanbot, line 94: Location: app/src/main/java/com/abk/kernel/ui/components/AppBackgroundHost.kt:66

Issue: The new `Image(painter = backgroundPainter, ...)` call omits `contentScale = ContentScale.Crop`. `Image` defaults to `Fit`, so wallpapers that do not match the screen aspect ratio will now letterbox/stretch differently than before. That also breaks the alignment assumptions used by the blur backdrops and card sampling code, so the frosted regions can sample the wrong part of the wallpaper.

Suggestion: Render the shared background image with `contentScale = ContentScale.Crop`, matching the previous `AsyncImage` behavior and the crop math used by the blur pipeline.

```suggestion
Image(
    painter = backgroundPainter,
    contentDescription = null,
    contentScale = ContentScale.Crop,
    modifier = Modifier.fillMaxSize()
)
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The blur integration is close, but the shared background rendering path in `AppBackgroundHost` introduces two regressions that affect wallpaper rendering and device stability.

### Decision
request_changes

### Notes

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
No actionable issues found in the PR diff. Static review and diff checks passed; Kotlin compilation could not be run because the repository has no gradle wrapper at ./gradlew.

### Decision
request_changes

### Notes
Codex recommendation: approve
This PR has 1826 changed lines, so policy requires request changes.

app/src/main/java/com/abk/kernel/ui/components/AppBackgroundHost.kt
```diff
@@ -17,28 +29,60 @@ fun AppBackgroundHost(
...
+    blurBackgroundEnabled: Boolean = false,
     content: @Composable () -> Unit
 ) {
     val hasBackground = backgroundEnabled && !backgroundUri.isNullOrBlank()
     val colorScheme = MaterialTheme.colorScheme
+    var backgroundCoordinates by remember { mutableStateOf<LayoutCoordinates?>(null) }
+    // One shared painter for the visible background image and every blur backdrop.
+    // Coil's AsyncImagePainter sizes its request from the onDraw viewport (fillMaxSize),
+    // so a large user wallpaper is decoded at screen resolution, not its original size.
+    val backgroundPainter = if (hasBackground) {
```
xingguangcuicanbot, line 51: Location: app/src/main/java/com/abk/kernel/ui/components/AppBackgroundHost.kt:42

Issue: `rememberAsyncImagePainter` does not automatically resolve the composable's layout constraints like `AsyncImage` does, so this change can decode the user's wallpaper at its original resolution. Large photos may therefore cause excessive memory use or OOM, especially while the separate downsampled blur bitmap is also retained.

Suggestion: Use a constraints-based size resolver for the shared painter, or retain `AsyncImage` for the visible background and provide a separately sized painter/request for blur sampling.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The shared background painter can regress memory usage by loading full-resolution custom images instead of sizing requests to the screen.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/screens/SettingsScreen.kt
```diff
@@ -583,9 +607,9 @@ fun SettingsScreen(
```
xingguangcuicanbot, line 582: Location: app/src/main/java/com/abk/kernel/ui/screens/SettingsScreen.kt:642

Issue: The SUSFS child page was migrated to `BlurScreenScaffold`, but its `ExpressiveTopBar` does not pass `enableBlur = state.blurEnabled`, unlike the other migrated screens. As a result, enabling Blur leaves this page's top bar opaque and inconsistent with the rest of the UI.

Suggestion: Pass `enableBlur = state.blurEnabled` to the SUSFS page's `ExpressiveTopBar`.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
Found one UI regression: the SUSFS settings page does not enable the new top-bar blur. Gradle verification could not run because this checkout does not contain a Gradle wrapper.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt
```diff
@@ -0,0 +1,557 @@
...
+    val context = LocalContext.current
+    // Restarting this effect on every size change cancels the in-flight debounce, so a
+    // rotation / split-screen resize only runs the final decode + StackBlur pass once the
+    // viewport settles. The result goes into the module-level cache, so transient changes
+    // (soft keyboard resizing the window) keep showing the previous frosted backdrop until
+    // the new one is ready, and matching viewports are restored from disk instead of
+    // re-blurring from scratch.
+    LaunchedEffect(uri, enabled, widthPx, heightPx) {
+        // Wallpaper changed: drop the stale memory + disk blur for the old image.
+        if (cachedBlurredCardUri != uri) {
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt:145

Issue: The persisted cache is deleted before it is ever read on every process start. `cachedBlurredCardUri` is initialized to `null`, so the first `LaunchedEffect` always enters the URI-change branch and deletes `abk_blurred_custom_background.jpg`; `loadBlurCache` later in the same effect can therefore never reuse the blur generated by a previous launch. This defeats the advertised on-disk cache and forces an image decode plus StackBlur every launch.

Suggestion: Only invalidate the disk cache when switching away from a URI already known in this process; then set the current URI before attempting to load the matching cache.

```suggestion
if (cachedBlurredCardUri != null && cachedBlurredCardUri != uri) {
    cachedBlurredCardBackground = null
    blurCacheFile(context).delete()
}
cachedBlurredCardUri = uri
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The on-disk blurred-background cache is unconditionally invalidated at startup, so its intended cross-launch reuse never occurs.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt
```diff
@@ -0,0 +1,582 @@
...
+ */
+private var cachedBlurredCardBackground by mutableStateOf<BlurredCardBackground?>(null)
+private var cachedBlurredCardUri by mutableStateOf<String?>(null)
+
+private fun blurCacheFile(context: android.content.Context, uri: String): File {
+    // Key the cache file by the background URI. String.hashCode() is spec-defined and
+    // stable across processes, so a previous launch's blur is only reused for the same
+    // wallpaper. Without this, a single-slot file would serve a stale wallpaper's blur
+    // after the background changes across launches.
+    val key = Integer.toHexString(uri.hashCode())
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt:95

Issue: The persistent cache filename is derived only from `String.hashCode()`. Different background URIs can legally produce the same 32-bit hash, causing the second URI to load and display the first URI's blurred bitmap instead of reprocessing its own image. This is a correctness issue in the new disk-cache path.

Suggestion: Use a collision-resistant cache key, such as a SHA-256 digest of the URI, rather than `String.hashCode()`. Also consider including the blur/downsample version in the filename so algorithm changes cannot reuse incompatible data.

```suggestion
val key = uri.toByteArray().contentHashCode().toUInt().toString(16)
```

app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt
```diff
@@ -0,0 +1,582 @@
...
+    // the new one is ready, and matching viewports are restored from disk instead of
+    // re-blurring from scratch.
+    LaunchedEffect(uri, enabled, widthPx, heightPx) {
+        // Wallpaper changed: drop the stale in-memory blur for the old image and clean up
+        // its on-disk cache. On a fresh process cachedBlurredCardUri is null, so the
+        // current wallpaper's disk cache is left intact for loadBlurCache below to reuse.
+        val previousUri = cachedBlurredCardUri
+        if (previousUri != null && previousUri != uri) {
+            cachedBlurredCardBackground = null
+            blurCacheFile(context, previousUri).delete()
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt:169

Issue: The cache cleanup deletes the previous URI's on-disk cache whenever the current URI changes. That makes the cache effectively single-use across wallpaper changes: switching back to a previously processed background always forces another decode and StackBlur pass, despite the implementation and comments describing a URI-keyed multi-entry cache.

Suggestion: Do not delete the previous URI's cache on every URI change. If storage reclamation is required, prune old cache files with an explicit size/count policy instead.

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The PR has two actionable cache issues: URI hash collisions can display the wrong blurred background, and URI changes unnecessarily delete reusable cache entries. The Gradle unit-test command could not be run because this checkout has no `gradlew` wrapper.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt
```diff
@@ -0,0 +1,620 @@
...
+    val heightPx = viewportSize.height
+    if (!enabled || uri.isNullOrBlank() || widthPx <= 0 || heightPx <= 0) {
+        return null
+    }
+    val context = LocalContext.current
+    // Re-blur only on a WIDTH change (rotation / split-screen). A height-only change is
+    // the IME (adjustResize), where the taller cached bitmap still covers the visible
+    // region, so keying the effect on width (not height) avoids a full decode + StackBlur
+    // pass on every keyboard toggle while still re-blurring real viewport changes.
+    LaunchedEffect(uri, enabled, widthPx) {
```
xingguangcuicanbot: Location: app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt:192

Issue: The pre-blur effect is keyed only on `widthPx`. That avoids IME-triggered reloads, but it also means real height-only viewport changes (for example freeform window resizing, top/bottom split-screen, or some foldable posture changes) never regenerate the blurred wallpaper. In those cases the cached bitmap keeps the old height and cards will sample/stretch the wrong region until the process is recreated.

Suggestion: Recompute the blurred backdrop when either viewport dimension changes, or add a stricter distinction between transient IME resizes and real window-size changes before skipping the reload.

```suggestion
    LaunchedEffect(uri, enabled, widthPx, heightPx) {
```

app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt
```diff
@@ -0,0 +1,620 @@
...
+                        // so the cancelled effect cannot later overwrite the newer effect's result.
+                        throw e
+                    } catch (e: Exception) {
+                        // Coil/StackBlur failure: fall back to the opaque surface tint. Errors
+                        // (e.g. OutOfMemoryError) are deliberately not caught here.
+                        null
+                    }
+                }
+        }
+        if (loaded != null) {
```
xingguangcuicanbot, line 248: Location: app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt:245

Issue: A canceled blur job can still win the race and publish stale output. `loader.execute(...)` and the StackBlur loop are synchronous CPU work, so cancellation is not observed until after `withContext(Dispatchers.Default)` returns. If the user changes the wallpaper or the viewport while a blur pass is in flight, the older coroutine can still reach this block and overwrite `cachedBlurredCardBackground` (and the cache file) with an outdated image.

Suggestion: Before mutating the shared cache, verify that the effect is still active and still corresponds to the current URI/viewport request.

```suggestion
        if (loaded != null) {
            if (cachedBlurredCardUri != uri || !kotlinx.coroutines.currentCoroutineContext().isActive) {
                return@LaunchedEffect
            }
            cachedBlurredCardBackground = loaded
            // JPEG compress + disk write; keep off the main thread (StrictMode / jank).
            withContext(Dispatchers.IO) {
                saveBlurCache(context, uri, loaded)
                pruneBlurCache(context)
            }
        }
```

xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
The blur integration is generally well-structured, but I found two correctness issues in the new card-backdrop pipeline that can leave the UI showing the wrong blurred image under common resize/background-change races.

### Decision
request_changes

### Notes

app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt
fanziyun: Fixed in 4e2c829e, keyed differently than the suggestion. Keying on both dimensions alone re-introduces a full decode + StackBlur on every IME open/close (the reason it was width-keyed). Instead the effect keys on both dimensions and skips the reload only while the cached bitmap still covers the new viewport (cached.width >= target.width and cached.height >= target.height). A height or width growth (freeform taller, fold unfold, rotation) no longer covers and re-blurs; shrink-only changes (IME) still cover and skip. The strip draw uses the same coverage condition so a stale bitmap is never stretched across grown bounds.

app/src/main/java/com/abk/kernel/ui/blur/BlurredCardBackdrop.kt
fanziyun: Not a bug. loader.execute(request) is a Coil suspend call, not synchronous CPU work, and withContext(Dispatchers.Default) re-checks cancellation when it resumes: if the LaunchedEffect was cancelled by a uri/viewport key change, withContext throws CancellationException instead of returning loaded, so cachedBlurredCardBackground = loaded is never reached. There is no suspension point between the loaded assignment and the cache write. The existing catch(CancellationException) { throw e } plus the next effect clearing the cache on URI change already covers this, so an isActive guard would be dead code.

PR #222, opened 2026-08-13
feat(ui): Add Miuix blur and custom card backdrop blur support

fanziyun:
## Summary

Ports **Miuix blur** (frosted-glass top / bottom bars) into the Expressive UI and adds an optional **custom-background card/tile blur** (frosted cards over the wallpaper), including a software fallback for pre-API-33 devices.

## What's included

**1. Miuix blur for the app chrome** (`ui/blur/` — `BlurBackdrop`, `BlurHost`, `BlurState`, `BlurScreenScaffold`, `BlurBackgroundPainter`)

- ~20 screens migrated from `Scaffold` to `BlurScreenScaffold`.
- The master blur switch is shown on **all** devices and is decoupled from device capability: the software card-blur path does not require runtime shaders, so it works on API 26-32 too.

**2. Custom card/tile backdrop blur** (`BlurredCardBackdrop.kt`)

- The custom background is pre-blurred once (downsampled 4x, StackBlur radius 45) and the matching sub-rect is drawn under each frosted card.
- Uses a software StackBlur fallback on pre-API-33 devices (no `RuntimeShader` dependency).

## Design decisions (and how prior review concerns are addressed)

- **minSdk-33 AAR on an API-26 app** (`tools:overrideLibrary` in the manifest): every path into the library is gated on `isRuntimeShaderSupported()` (API 33+), and the AAR registers no `ContentProvider` / initializer, so nothing library-side executes on API 26-32. This is documented in the manifest comment.
- **Decode sizing**: the shared wallpaper painter is sized to the display (`ImageRequest.size(displayMetrics)`), and the card-blur decode is constrained to `viewport/4` (`ImageRequest.size(...)`) with `allowHardware(false)` — never a full-resolution decode.
- **Memory / bitmaps**: blur runs on `Dispatchers.Default`; the JPEG cache write runs on `Dispatchers.IO`; intermediate bitmaps are recycled; `ContentScale.Crop` is applied everywhere.
- **On-disk blur cache**: keyed by `SHA-256(uri)` (collision-resistant, not `hashCode()`), with a version + radius + downsample component so a future algorithm change cannot silently reuse stale data. Entries are **count-pruned** (keep 4 newest) rather than deleted on switch, so toggling between wallpapers reuses the cache. Corrupt files are deleted on decode failure; orphaned `.tmp` files are pruned on save.
- **Viewport resize**: the anchor size is mirrored into snapshot state (`LocalBlurBackgroundSize`, because `LayoutCoordinates.size` is not observable). Re-blur is driven by a **coverage check** — the cached bitmap is re-blurred only when the viewport grows past it (rotation, split-screen, freeform taller, fold unfold), and skipped for shrink-only changes (IME), where the taller bitmap still samples the correct region. This avoids both stale-stretched backdrops and per-keypress re-blurs.
- **IME handling**: `windowSoftInputMode="adjustResize"` (matches `dev`), so the keyboard never covers content on Android 11+; the coverage check ensures an IME height change does not re-run the blur.
- **No first-frame flash**: `LocalBlurredCardBackgroundEnabled` is a synchronous signal, so cards stay translucent from the first frame instead of flashing opaque→translucent when the frosted backdrop arrives.

## Testing

- `BlurConfigTest` unit-tests the blur config → enabled-state mapping.
- The resize / IME / cache-reuse paths are exercised on the emulator and a physical API-26 device (software fallback).

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -201,6 +201,9 @@ dependencies {
     // Preferences
     implementation(libs.datastore.preferences)
 
+    // Blur / glass
+    implementation(libs.miuix.blur)
+
     testImplementation(libs.junit)
     testImplementation(libs.robolectric)
     testImplementation(libs.kotlinx.coroutines.test)
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -42,7 +42,7 @@
             android:name=".MainActivity"
             android:exported="true"
             android:launchMode="singleTop"
-            android:windowSoftInputMode="adjustNothing"
+            android:windowSoftInputMode="adjustResize"
             android:theme="@style/Theme.ABK">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
@@ -91,7 +91,7 @@
         <activity
             android:name=".extensions.AbkExtensionManagerActivity"
             android:exported="false"
-            android:windowSoftInputMode="adjustNothing"
+            android:windowSoftInputMode="adjustResize"
             android:theme="@style/Theme.ABK" />
 
         <service
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -15,7 +15,14 @@
     <uses-permission android:name="android.permission.QUERY_ALL_PACKAGES" />
     <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
 
-    <!-- Runtime shader blur is guarded in code while the app still supports API 26+. -->
+    <!--
+      The miuix-blur AAR declares minSdk 33 while this app supports API 26+. The override
+      keeps the build working; safety is enforced at runtime. Every path into the library
+      (rememberBlurBackdrop, blurSource, blurEffect, blurSourceBody, isBlurActive) is gated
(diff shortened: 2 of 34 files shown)


xingguangcuicanbot (changes requested):
## Automated PR Review

### Summary
I did not find any actionable correctness, security, regression, or maintainability issues in the visible `origin/dev...HEAD` diff. The blur gating, fallback paths, scaffold migration, and preference plumbing are internally consistent. Validation was limited to static review because this checkout does not include a Gradle wrapper and no system `gradle` binary is available here.

### Decision
request_changes

### Notes
Codex recommendation: approve
This PR has 2024 changed lines, so policy requires request changes.

xingguangcuicanbot (dismissed):
## Automated PR Review

### Summary
I reviewed the checked-out PR against `origin/dev...HEAD`, including the new blur infrastructure, cache/bitmap lifecycle, API-33 gating, manifest/build changes, and the migrated screen scaffolds. I did not find any distinct actionable issues in the visible diff from a read-only static review. Residual risk remains around runtime-only UI/inset behavior on device classes that cannot be exercised here.

### Decision
request_changes

### Notes
Codex recommendation: approve
This PR has 1968 changed lines, so policy requires request changes.

xingguangcuicanbot (approved):

